PR #14, opened 2023-04-14
End-to-end testing workflows for both apps

barbarah:
I'm using the GitHub Action `patrickedqvist/wait-for-vercel-preview`. Because with this action, I can set an environment.

I also tried the following, but it did NOT work because it would always skip the workflow completely if the other app first triggered a successful deployment.

```
name: End-to-end testing Dataset Browser
on: deployment_status
jobs:
  run-end-to-end-tests:
    if: github.event.deployment_status.state == 'success' && github.event.deployment_status.environment == 'Preview – colonial-collections-dataset-browser' 
    .....
````

### CODE REVIEW

File: .github/workflows/end-to-end-testing-dataset-browser.yml [hunk 1]
```diff
@@ -0,0 +1,28 @@
...
+    branches: ["main"]
+    paths:
+      - 'package.json'
+      - 'apps/dataset-browser/**'
+      - .github/workflows/end-to-end-testing-dataset-browser.yml
+jobs:
+  run-end-to-end-tests:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Waiting for 200 from the Vercel Preview
```
sdevalk, line 13: Minor:
1. 'Wait' (to be consistent with the other names)?
2. What is the meaning of 200? Wait for 200 milliseconds? Can we make this explicit?

sdevalk (approved):
Perfect!

PR #16, opened 2023-04-14
Added error pages with end-to-end tests

barbarah:
The content pages aren't working. Before fixing the pages, I want CI to tell me that the pages are broken.

I have:

- Added error pages
- Moved the layout to the root level, so the error page looks nice + this will fix the problem in this pull request: https://github.com/colonial-heritage/colonial-collections/pull/3
- Added tests to check all pages
- Added a script to the package.json's to run all cypress tests easily

Notes:

- The error page cannot be translated, so I just added an English text.
- This pull request contains a lot of double code. But I don't want to move the layout, error and not-found page to the UI page without discussing the base design of the research app first.
- The Cypress tests will fail; this is good because the content pages have a bug.
- If you want to see the failing tests, this pull request needs to be merged: https://github.com/colonial-heritage/colonial-collections/pull/14


apps/dataset-browser/src/app/error.tsx
```diff
@@ -0,0 +1,15 @@
+'use client'; // Error components must be Client components
+
+import {useEffect} from 'react';
+
+export default function Error({error}: {error: Error}) {
+  useEffect(() => {
```
sdevalk, line 6: Curious: do we need a `useEffect` here? What if we console.log the error without it?

barbarah: I got this from the Next.js docs: https://beta.nextjs.org/docs/routing/error-handling

I think it's to prevent multiple console logs if there are any rerenders. Maybe from the layout. But in our case, I don't think it's necessary. So I will remove it.

PR #28, opened 2023-04-18
feat: iris

sdevalk:
This PR extracts the IRI helpers from the Dataset Browser and moves them to their own package. This allows for re-use of the helpers in the Research App.


barbarah (approved):
Nice!

PR #37, opened 2023-04-20
Move client list logic to zustand store

barbarah:
This pull request is the start of the shared filter logic for the Dataset Browser and the Researcher.

- I have created a Zustand store in the new 'list-store' package.
- I have refactored to client component, so it uses the list store and removed dataset-specific logic (mainly the filters).
- I have moved one function of `lib/search-params` (getUrlWithSearchParams) to the store. The other functions are more project-specific and may need to be moved, but this takes more time.

Nest steps, out of scope for this pull request:

1. Create the client filters for the Researcher with duplicate UI components first.
2. I want to refactor the client-filter component and his children and move to the UI package.
     - Use the store is all components instead of passing down the state. Logic can also move to the store, leaving only UI in the client components.
     - Change to compound components with dot notation. With compound components, it is easier to have UI differences in the apps.


apps/dataset-browser/src/app/[locale]/store-initializer.tsx
```diff
@@ -0,0 +1,36 @@
...
+  offset,
+  limit,
+  query,
+  sortBy,
+  selectedFilters,
+}: Props) {
+  const initialized = useRef(false);
+  if (!initialized.current) {
+    useListStore.setState({
+      totalCount,
```
Comment (barbarah) [inline apps/dataset-browser/src/app/[locale]/store-initializer.tsx:26]: Using this code for the researcher, I realized that `totalCount` won't update after initialization. The client also updates the other properties. But `totalCount` only comes from the server.

packages/list-store/src/search-params.ts
```diff
@@ -0,0 +1,65 @@
...
+    .number()
+    .default(0)
+    // Don't add the default offset of 0 to the search params
+    .transform(offset => (offset > 0 ? `${offset}` : '')),
+  sortBy: z
+    .nativeEnum(SortBy)
+    .default(defaultSortBy)
+    // Don't add the default sort to the search params
+    .transform(sortBy => {
+      console.log(sortBy, defaultSortBy);
```
barbarah, line 21: There is still some debug code here.

barbarah:
There is some infinite loop going on. Not sure where this is coming from.

PR #46, opened 2023-04-26
new mock data for Dataset Browser

sdevalk:
This PR adds some mock data for the Dataset Browser. The data will allow us to present transparency measurements of datasets in the app. The Dataset Fetcher doesn't make use of this data yet - I still need to update it.


barbarah (approved):
Looking forward to working with this new data!

PR #50, opened 2023-04-30
feat: dataset measurements

sdevalk:
This PR adds the property '[measurements](https://github.com/colonial-heritage/colonial-collections/blob/feat-dataset-measurements/apps/dataset-browser/src/lib/datasets/fetcher.integration.test.ts#L1554-L1609)' to the `Dataset`. The measurements contain the results of a transparency analysis, to find out how transparent a dataset is. The current measurements are very basic - these will have to be expanded at some point. The measurements are meant to get a conversation going with users of the Dataset Browser.

Apart from several refactoring changes, this PR also brings in another change: I have renamed the `dataset-fetcher` folder to `datasets` because the folder isn't only about 'fetching' anymore (but also about 'enriching' datasets with information, currently with the measurements).


sdevalk:
@barbarah I've updated the PR: the metrics now use fixed IDs (IRIs):

1. https://data.colonialcollections.nl/metrics#open-license
1. https://data.colonialcollections.nl/metrics#languages
1. https://data.colonialcollections.nl/metrics#online
1. https://data.colonialcollections.nl/metrics#structured-format
1. https://data.colonialcollections.nl/metrics#open-format
1. https://data.colonialcollections.nl/metrics#rdf-format

You can use these in the frontend, e.g. for adding localized labels to the metrics.

The IDs will probably change at some point (due to new insights or new metrics), but for now they're stable.

barbarah (approved):

PR #57, opened 2023-05-03
feat: identifier

sdevalk:
This PR expands the 'object fetcher' of the researcher: it introduces the new, required property `identifier`. This property can be used for searching and presenting the ID that the owner has assigned to the object (e.g. a catalogue number or a record identifier).


barbarah (approved):
Nice, I see the preview of this branch is working correctly.

PR #71, opened 2023-05-10
feat: person fetcher

sdevalk:
This PR introduces a new entity type: `Person`. It adds a separate `PersonFetcher`, next to the `HeritageObjectFetcher`. We can use this new fetcher for searching for and presenting information about persons in their own result view and detail page.

The code is 'good enough' - it works :) But there are several questions in the code about the information that we have about persons - these need to be resolved soon (depending on the data from the data providers and the frontend design that needs to be implemented).


barbarah (approved):
Looks good and is simular to the other fetchers, so easy to use.

sdevalk:
> Looks good and is simular to the other fetchers, so easy to use.

Thanks, perfect!

I still need to add a facet for 'Dataset'. I'll create a separate PR for this.

Issue #80, opened 2023-05-10
Add the possibility to share objects in the Research app


barbarah:

barbarah:
@Doppen, The URL to the objects, are already shareable. So this could be a simple button that copies the URL, and a message "The URL is copied". Is that what you want with this button? Or is it more complex?

Issue #76, opened 2023-05-10
Show related persons on the details page of the Research app


barbarah:

barbarah:
@sdevalk and @Sabachi I think you two need to discuss if this can be possible.

Issue #75, opened 2023-05-10
Extend the licence information with an image and more info


barbarah:
See design:

![image](https://github.com/colonial-heritage/colonial-collections/assets/1481602/a56b8291-1bf8-404a-8077-68d5fff2af2a)

barbarah:
Is this always the license of the dataset? If so, I could collect the dataset. However, 
@sdevalk, this means the data fetcher should be a separate package. 


I also do not have an image; I do not know how we can get the image.

Issue #74, opened 2023-05-10
Extend the owners information with an image and contact details


barbarah:
![image](https://github.com/colonial-heritage/colonial-collections/assets/1481602/75a9fab5-f9d9-40ce-82b4-a146e398f183)

barbarah:
The object heritage fetcher returns an owner ID and name. We do not have an image and contact details.

@Doppen en @sdevalk  What is the object ID? We have an URL, but not something in the format of the design.

If the owners are a fixed list. We could add a JSON to the code base with the missing data. OR we could add it to the API. @sdevalk, any thoughts on this?

Doppen:
@barbarah The Object ID is the ID of the object from the museum self. See https://collectie.wereldculturen.nl/#/query/c5d7ab75-c6da-4812-9094-819d703617ee

PR #82, opened 2023-05-10
Added a details page for objects

barbarah:
This is the first version of the details page. 

@Doppen Can you look at the design? The images now have a fixed height. But this does not work well with portrait images. Can you change the layout in the code OR make three versions in Adobe xd? With one version with an image in portrait, one in landscape and one square image. Also, the background is now grey. In the XD version, it's white. But this is also for the homepage so I will change this in a separate ticket.

I created these related tickets:

- #72 
- #73
- #74 
- #75
- #76
- #77
- #78
- #79
- #80
- #83

### DIFF
--- a/apps/researcher/cypress/e2e/object.cy.ts
+++ b/apps/researcher/cypress/e2e/object.cy.ts
@@ -1,5 +1,5 @@
 describe('Object details page', () => {
-  it('loads an existing object', () => {
+  it('opens the object page if clicked on in the search list', () => {
     cy.visit('/en');
     // Get the name of the first object in the list.
     cy.getBySel('object-card-name')
@@ -11,14 +11,14 @@ describe('Object details page', () => {
         cy.location('pathname', {timeout: 60000}).should('include', '/object');
         // On the details page.
         cy.getBySel('error').should('not.exist');
+        cy.getBySel('no-object').should('not.exist');
         cy.getBySel('page-title').then($detailsName => {
           expect($cardName.text()).equal($detailsName.text());
-          cy.getBySel('no-object').should('not.exist');
         });
       });
   });
 
-  it('shows an error message if no object can be found', () => {
+  it('shows an error message if no object matches the ID', () => {
     cy.visit('/en/object/anIdThatDoesNotExist');
     cy.getBySel('no-object').should('exist');
     cy.getBySel('object-name').should('not.exist');
--- a/apps/researcher/cypress/e2e/object.cy.ts
+++ b/apps/researcher/cypress/e2e/object.cy.ts
@@ -1,0 +1,26 @@
+describe('Object details page', () => {
+  it('loads an existing object', () => {
+    cy.visit('/en');
+    // Get the name of the first object in the list.
+    cy.getBySel('object-card-name')
+      .first()
+      .then($cardName => {
+        // Navigate to the first object details page.
+        cy.getBySel('object-card-name').first().click();
+        // Wait for the page to load.
+        cy.location('pathname', {timeout: 60000}).should('include', '/object');
+        // On the details page.
+        cy.getBySel('error').should('not.exist');
+        cy.getBySel('page-title').then($detailsName => {
+          expect($cardName.text()).equal($detailsName.text());
+          cy.getBySel('no-object').should('not.exist');
+        });
+      });
+  });
+
+  it('shows an error message if no object can be found', () => {
+    cy.visit('/en/object/anIdThatDoesNotExist');
+    cy.getBySel('no-object').should('exist');
+    cy.getBySel('object-name').should('not.exist');
+  });
+});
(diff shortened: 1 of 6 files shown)


apps/researcher/cypress/e2e/object.cy.ts
```diff
@@ -0,0 +1,26 @@
...
+        // On the details page.
+        cy.getBySel('error').should('not.exist');
+        cy.getBySel('page-title').then($detailsName => {
+          expect($cardName.text()).equal($detailsName.text());
+          cy.getBySel('no-object').should('not.exist');
+        });
+      });
+  });
+
+  it('shows an error message if no object can be found', () => {
```
sdevalk, line 21: 'no object', I think, doesn't make clear that a specific object is requested. Perhaps: `shows an error message if no object matches the ID` or `shows an error message if the object is not found`?

apps/researcher/src/app/[locale]/object/[id]/page.tsx
```diff
@@ -0,0 +1,185 @@
...
+function InfoBlock({object, objectKey, singleLine}: InfoBlockProps) {
+  const t = useTranslations('Details');
+
+  const items = object[objectKey];
+
+  if (!items || !items.length) {
+    return null;
+  }
+
+  const itemStrings = items.map(item =>
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/object/[id]/page.tsx:58]: Or: `itemValues` (so as not to limit it to strings)?

apps/researcher/src/app/[locale]/object/[id]/page.tsx
```diff
@@ -0,0 +1,185 @@
...
+              className="rounded-md bg-white px-3 py-2 text-sm font-semibold text-gray-900 shadow-sm ring-1 ring-inset ring-gray-300 hover:bg-gray-50"
+            >
+              {t('shareButton')}
+            </button>
+          </div>
+          {object.images?.length && (
+            <div className="relative col-span-2 h-[250px] w-full">
+              <Image
+                alt={object.name || object.id}
+                src={object.images[0].contentUrl}
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/object/[id]/page.tsx:163]: What should happen if an object has multiple images? The order of which the images are returned by the backend is undefined - we don't know whether the first one is the right/best one. @Doppen What do you think?

apps/researcher/src/messages/nl/messages.json
```diff
@@ -20,5 +20,25 @@
...
     "HeritageObject": "Erfgoedobject"
+  },
+  "Details": {
+    "backButton": "Terug naar resultaten",
+    "notImplemented": "Niet geïmplementeerd",
+    "noData": "Geen gegevens",
+    "description": "Beschrijving",
+    "date": "Datum",
+    "owner": "Eigenaar",
+    "handle": "Hendel",
```
sdevalk, line 31: This should be `Handle`, too - in English :) I don't think we should implement this, by the way - a Handle is a type of persistent identifier, but there are more. Not every data providers uses Handles. But let's first deploy this and then discuss :)

sdevalk (commented):
Nice!

apps/researcher/cypress/e2e/object.cy.ts
```diff
@@ -0,0 +1,26 @@
+describe('Object details page', () => {
+  it('loads an existing object', () => {
```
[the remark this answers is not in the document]
barbarah: I changed it to "opens the object page if clicked on in the search list".

apps/researcher/src/app/[locale]/object/[id]/page.tsx
Comment (barbarah) [reply apps/researcher/src/app/[locale]/object/[id]/page.tsx:163]: Good question. We probably need an image gallery. [This](https://tailwindui.com/components/ecommerce/components/product-overviews#component-13a89c2dc50a31afd66541dc28fa3c13) for example.

I suggest adding a ticket for multiple images. So we can tackle this as a separate feature.

Comment (barbarah) [reply apps/researcher/src/app/[locale]/object/[id]/page.tsx:163]: https://github.com/colonial-heritage/colonial-collections/issues/new

Comment (barbarah) [reply apps/researcher/src/app/[locale]/object/[id]/page.tsx:163]: This question is also relevant for the object car (#81).

apps/researcher/src/app/[locale]/object/[id]/page.tsx
```diff
@@ -0,0 +1,185 @@
...
+          <div>
+            <H3>{t('handle')}</H3>
+            <NotImplemented />
+          </div>
+          <div>
+            <H3>{t('license')}</H3>
+            <NotImplemented />
+          </div>
+          <div className="space-y-4">
+            <H2>{t('moreInfo')}</H2>
```
Comment (barbarah) [inline apps/researcher/src/app/[locale]/object/[id]/page.tsx:126]: @Doppen For your information:  Based on the design, I have added single-line (joined with a ',') and multi-lined info blocks. The sub-headers are different from the designs.  So I guessed which ones were single- and which were multi-lined. Since we don't have real data, it's probably also hard for you to decide which one to use where. So I suggest we just start with these values and check back when we have more data.

sdevalk (approved):

PR #89, opened 2023-05-11
Added the content pages to both apps

barbarah:
Both apps now use the same content package. So the content both apps use only needs to be managed in one place.

I also added the missing screen reader menu for the research app.

The `layout.tsx` and `navigation.tsx` for both apps are now exactly the same. I have not moved the files to the UI package because I don't know if they are going to stay exactly the same.

The content needs to be updated. For example, the page title of the about page is "About the Dataset Browser".


sdevalk (approved):
Perfect!

PR #91, opened 2023-05-12
Refacor list pages: Extract client components and move to UI package

barbarah:
This is the second part of the refactor needed for a third list page (persons).

I have moved all parts of `client-filters.tsx` that actually needs to be a client component to the UI package. What's left can be a server component, so I have moved it to the `page.tsx`.


apps/dataset-browser/src/app/[locale]/page.tsx
```diff
@@ -19,27 +38,53 @@ const filterKeysOrder: ReadonlyArray<keyof SearchResult['filters']> = [
...
   const sortBy = getClientSortBy({
     sortBy: searchOptions.sortBy,
     sortOrder: searchOptions.sortOrder,
   });
 
-  let hasError, searchResult;
+  let hasError, searchResultOrUndefined;
   try {
-    searchResult = await datasetFetcher.search(searchOptions);
+    searchResultOrUndefined = await datasetFetcher.search(searchOptions);
```
Comment (sdevalk) [inline apps/dataset-browser/src/app/[locale]/page.tsx:81]: The name of the variable suggests there is a 'search result' or an 'undefined'. But if I look at the return value of `search()`, there can only be a 'search result'?

apps/dataset-browser/src/app/[locale]/page.tsx
```diff
@@ -19,27 +38,53 @@ const filterKeysOrder: ReadonlyArray<keyof SearchResult['filters']> = [
...
   } catch (error) {
     hasError = true;
     console.error(error);
   }
+  const searchResult = searchResultOrUndefined as SearchResult;
```
Comment (sdevalk) [inline apps/dataset-browser/src/app/[locale]/page.tsx:86]: To me it looks a bit strange to convert a `searchResultOrUndefined` to a `SearchResult` - as if you're ignoring the `undefined`. Is this conversion necessary?

apps/researcher/src/app/[locale]/page.tsx
```diff
@@ -18,27 +36,53 @@ const filterKeysOrder: ReadonlyArray<keyof SearchResult['filters']> = [
   'subjects',
 ];
 
+export interface FacetMenuProps {
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/page.tsx:39]: To be sure: is the `export` required?

packages/ui/list/paginator.tsx
```diff
@@ -10,6 +10,10 @@ export function Paginator() {
   const end = endMax < totalCount ? endMax : totalCount;
   const t = useTranslations('Paginator');
 
+  if (!totalCount || totalCount === 0) {
```
sdevalk, line 13: If `totalCount` = 0, it succeeds on the first check (`!totalCount`) and not get to the second - right? If the second is import, rephrase the first check (e.g. `(totalCount === undefined)`)?

sdevalk (commented):
Nice! I've added some minor questions.

apps/dataset-browser/src/app/[locale]/page.tsx
Comment (barbarah) [reply apps/dataset-browser/src/app/[locale]/page.tsx:86]: I had some weird type errors within the map on line 165. Outside the map Typescript understood that the type of `searchResult` was `SearchResult`. But within the map, the type of `searchResult` was `any`. I changed the code a bit, so I don't need the `searchResultOrUndefined` variable.

sdevalk (approved):
👍🏻

PR #100, opened 2023-05-18
Person card

barbarah:
This is the person card in the person list [based on the design](https://xd.adobe.com/view/8663c916-07b9-4b14-a63a-bc1855033d21-d330/). I made the background white instead of grey because the main background is still grey.

### DIFF
--- a/apps/researcher/src/messages/nl/messages.json
+++ b/apps/researcher/src/messages/nl/messages.json
@@ -44,7 +44,7 @@
   },
   "PersonCard": {
     "person": "Persoon",
-    "relatedObjects": "Gerelateerde objecten onbekend (niet geïmplementeerd)",
+    "relatedObjects": "0 gerelateerde objecten (niet geïmplementeerd)",
     "birthPlaceUnknown": "Geboorteplaats onbekend",
     "birthYearUnknown": "Geboortejaar onbekend",
     "deathPlaceUnknown": "Plaats van overlijden onbekend",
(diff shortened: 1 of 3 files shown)


apps/researcher/src/messages/nl/messages.json
```diff
@@ -43,7 +43,12 @@
     "heritageObject": "Erfgoedobject"
   },
   "PersonCard": {
-    "person": "Persoon"
+    "person": "Persoon",
+    "relatedPersons": "Hoeveelheid gerelateerde objecten Onbekend (niet geïmplementeerd)",
```
sdevalk, line 47: Is this indeed about a number ('hoeveelheid')? Or simply: 'Onbekend (niet geïmplementeerd)'?

barbarah: Only 'Onbekend (niet geïmplementeerd)' will remove the context for the user. See image

image

I could do 'gerelateerde objecten Onbekend (niet geïmplementeerd)'. But I do think this section is about the amount of related objects.

I do see the key is wrong. It should be `relatedObjects`.

barbarah: If the word 'hoeveelheid' is a problem I will change it to 'Gerelateerde objecten onbekend (niet geïmplementeerd)'

barbarah: Thinking about this some more. I will add the amount. This will be a static 0. This will represent the expected behaviour better. I will keep the '(niet geïmplementeerd)' part, so it's clear this is always zero.

sdevalk (approved):
👍🏻

Issue #84, opened 2023-05-10
Only show images if the user wants to see the object images


barbarah:
Since users can have a slow internet connection, images should only be shown if the user wants to load images. This can be set with a toggle or at individual images.

PR #101, opened 2023-05-18
Add an image gallery to the object details page

barbarah:
Based on tailwindui: https://tailwindui.com/components/ecommerce/components/product-overviews#component-13a89c2dc50a31afd66541dc28fa3c13

Next step is #84

### DIFF
--- a/apps/researcher/src/app/[locale]/object/[id]/page.tsx
+++ b/apps/researcher/src/app/[locale]/object/[id]/page.tsx
@@ -92,7 +92,7 @@ export default async function Details({params}: Props) {
     object.images?.map((image, i) => ({
       ...image,
       src: image.contentUrl,
-      alt: `${object.name}, image - ${i}`,
+      alt: `${object.name} #${i + 1}`,
     })) ?? [];
 
   return (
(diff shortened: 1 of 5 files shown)


apps/researcher/src/app/[locale]/object/[id]/page.tsx
```diff
@@ -88,6 +88,13 @@ export default async function Details({params}: Props) {
     return <div data-testid="no-object">{t('noObject')}</div>;
   }
 
+  const imagesWithAlt =
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/object/[id]/page.tsx:91]: Minor: is it necessary to say 'withAlt'? If we, for some reason, decide to remove the alt, we also have to rename the variable. Is it possible to use a more persistent name?

apps/researcher/src/app/[locale]/object/[id]/page.tsx
```diff
@@ -88,6 +88,13 @@ export default async function Details({params}: Props) {
...
+    object.images?.map((image, i) => ({
+      ...image,
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/object/[id]/page.tsx:93]: This merges all properties of `image` (e.g. `id` and `contentUrl`). But according to the `Props` interface of `gallery.tsx` it should only contain `id`. Perhaps selectively merge properties?

sdevalk (approved):
👍🏻

PR #102, opened 2023-05-19
refactor: mock data blank nodes

sdevalk:
The mock data and the object and person fetchers assumed we would we working with IRIs for identifying terms, such as places, types, subjects, creators, etc. However, these IRIs are often missing in the source data of data providers. This PR changes the mock data and fetcher: they now use literals (e.g. names) instead of IRIs.


barbarah (approved):
Looks good. These changes will not affect the front end.

PR #108, opened 2023-05-20
fix: revalidate pages

sdevalk:
This PR sets the `revalidate` property of each page to 60 seconds. This makes the app a bit snappier because results are being cached. The `revalidate` gave some issues before (e.g. stale cache) - if these persist we can revert this change.


barbarah (approved):
This will definitely speed up navigating back and forth. And easy to roll back if there are still caching errors. So lets see if this works!

Issue #15, opened 2023-04-14
Build in facets loading


Doppen:

Doppen:
**New code for loading the results.**
Within the `h1` tag:
`<div class="animate-pulse opacity-90 text-base">Loading...</div>`

Code instead of the result cards:
```          <div class="group relative flex flex-col overflow-hidden drop-shadow-md bg-white min-h-[150px] animate-pulse opacity-100">
            <div class="flex flex-1 flex-col space-y-2 p-6">
              <div class="bg-sand-100 h-4 w-3/4 rounded"></div>
              <div class="bg-sand-100 h-4 w-1/2 rounded mt-12"></div>
            </div>
          </div>

          <div class="group relative flex flex-col overflow-hidden drop-shadow-md bg-white min-h-[150px] animate-pulse opacity-80">
            <div class="flex flex-1 flex-col space-y-2 p-6">
              <div class="bg-sand-100 h-4 w-3/4 rounded"></div>
            </div>
          </div>

          <div class="group relative flex flex-col overflow-hidden drop-shadow-md bg-white min-h-[150px] animate-pulse opacity-60">
            <div class="flex flex-1 flex-col space-y-2 p-6">
              <div class="bg-sand-100 h-4 w-1/4 rounded"></div>
            </div>
          </div>```

PR #123, opened 2023-05-31
Loading screen for the researcher

barbarah:
This is the first pull request about loading components. This pull request contains a loading screen for the researcher app. The following steps (out of the scope of this pull request) are the loading component after clicking on a filter and adding the loading screen to the dataset browser. It would be nice to finish this layout first so I can use it for the next steps.

@Doppen  I did not add "loading..." to the page header as suggested in the issue #15 . It felt busy. But you are more than welcome to play around with the loading component. I will wait for your approval before working on the loading after filtering.

It can be hard to see the loading screen, adding this locally to the page.tsx can help: `await new Promise(resolve => setTimeout(resolve, 2000));`

About the tabs: In the current production version of the researcher, after clicking on one tab, it will wait till the list is loaded before the related tab is active. I want the tab to be active sooner.

To fix this, I have added the Tabs component to the Next.js loading component. However, this did bring a new problem. Showing an active tab in a server component can be complex. See this issue: https://github.com/vercel/next.js/issues/43704.

That is why the Tabs component was a client component. But in the loading component, `usePathname` was undefined. So with this solution, the tabs were loading, but the active tab was only selected once the list was fully loaded. 

So I decided to go the server component route and select the active tab using headers and a regex, described in the nextjs issue: https://github.com/vercel/next.js/issues/43704

### DIFF
--- a/apps/researcher/src/app/[locale]/tabs.tsx
+++ b/apps/researcher/src/app/[locale]/tabs.tsx
@@ -1,16 +1,22 @@
-'use client';
-
 import classNames from 'classnames';
 import {ObjectIcon, PersonIcon} from '@/components/icons';
-import {usePathname} from 'next-intl/client';
 import {useTranslations} from 'next-intl';
+import {headers} from 'next/headers';
 
 export default function Tabs() {
   const t = useTranslations('Tabs');
-  const pathname = usePathname();
+
+  // The header 'x-invoke-path' always contains the language prefix even if the next-intl middleware removed it
+  const activePath = headers().get('x-invoke-path') || '';
+
   const tabs = [
-    {name: t('objects'), href: '/', icon: ObjectIcon},
-    {name: t('persons'), href: '/persons', icon: PersonIcon},
+    {name: t('objects'), current: /^\/[^/]+$/, href: '/', icon: ObjectIcon},
+    {
+      name: t('persons'),
+      current: /\/persons$/,
+      href: '/persons',
+      icon: PersonIcon,
+    },
   ];
 
   return (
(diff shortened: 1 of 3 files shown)


barbarah:
@sdevalk I removed you as a reviewer. The active tabs aren't working in the preview. It is locally. So I need to fix that first.

apps/researcher/src/app/[locale]/tabs.tsx
```diff
@@ -1,24 +1,37 @@
...
-  const pathname = usePathname();
+
+  const activePath = headers().get('x-pathname') || '/';
+
   const tabs = [
-    {name: t('objects'), href: '/', icon: ObjectIcon},
-    {name: t('persons'), href: '/persons', icon: PersonIcon},
+    {
+      name: t('objects'),
+      isCurrentRegex: `^/(${locales.join('|')})?$`,
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/tabs.tsx:15]: Could, at some point, the locales contain characters that have meaning in regular expressions? If so, this could will break. Escape the locales?

sdevalk (approved):
A creative solution! :) I've added one question.

apps/researcher/src/app/[locale]/tabs.tsx
Comment (barbarah) [reply apps/researcher/src/app/[locale]/tabs.tsx:15]: Good question. The locales are based on official language codes. Non contain characters that have meaning in regular expressions. So I will leave it like this.

https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes

Comment (sdevalk) [reply apps/researcher/src/app/[locale]/tabs.tsx:15]: Perhaps you could add a comment to the code? Then it's clear why the locale values aren't escaped?

PR #126, opened 2023-06-05
Remove `.vscode` folder in sub directories

barbarah:
There is already a  `.vscode` folder at the root level. Next.js automatically added these sub-folders. It's now possible to remove them: https://github.com/vercel/next.js/issues/42213


sdevalk (approved):
Nice!

PR #152, opened 2023-06-21
feat: prov info

sdevalk:
This PR adds provenance information to information about heritage objects in the Research App. It does so by introducing an 'enricher': it fetches the prov info from the SPARQL endpoint and adds it to the heritage objects.

Some remarks:
1. There are two known issues that I'll fix in separate PRs:
a. The provenance events of an object currently aren't sorted - their order is random.
b. The types of provenance events also show up in the 'Types' facet of the App (whereas this facet should only consist of the types of heritage objects)
3. The PR assumes that we only want to present the provenance information on the detail page of an heritage object, not in the search results. I can change this easily, though.
4. The PR makes a lof of assumptions about the way that the provenance information is structured. But so far I haven't seen real data yet. The code probably will have to be updated as soon as this data comes in.


sdevalk:
Superseded by #189 -- closing.

PR #158, opened 2023-07-03
Bump "next-intl" to "3.0.0-beta.7"

barbarah:
`next-intl` is not being updated by Depentabot because we use the beta version. I had some problems configuring Clerk in combination with `next-intl`. So I decided first to update "next-intl" before I started posting questions to the Clerk team.

Fix deprecation warnings.

- getTranslations is deprecated, use getTranslator
- The import for Link from next-intl has changed.

Fix bugs

- Set the middleware header correctly. We need to set a request header, not the response header. Not sure why it used to work, but it did not work after. See discussion https://github.com/amannn/next-intl/discussions/363
- Layout should be on [locale] level. I updated the layout based on the docs. It was not working on 'app' folder level.


barbarah:
@sdevalk I would like these changes to implement Cleck with the newest version. I don't think these changes are very complex, and the test servers are working correctly. So I'm going to merge without a review.

PR #164, opened 2023-07-07
Move the 'about' content back to apps

barbarah:
As discussed at the daily. Camiel would like to write app specific 'about' pages. This pull request only contains the move so that Camiel can update the content.


apps/researcher/src/messages/en/about.mdx
```diff
@@ -0,0 +1,27 @@
+# About the Dataset Browser
```
sdevalk, line 1: `About the Research Application`? But I guess Camiel can update this :)

sdevalk (approved):
👍🏻

apps/researcher/src/messages/en/about.mdx
barbarah: Yes. Currently, the text is also "About the Research Application" for both apps. I decided not to change any text in this pull request.

PR #165, opened 2023-07-10
Added Clerk for authentication

barbarah:
This pull request adds Cleck for authentication. This is just a basic implementation.

- With localization.
- With organizations (renamed to 'communities').
- For now, all paths are public.
- Logging in will only add a user en community button; there are no private components yet.
- The styling is the basic Cleck styling.
- The middleware had problems with the dots in the dynamic route segments. So I replaced the dots with '%2E'.


apps/researcher/cypress/e2e/home.cy.ts
```diff
@@ -1,14 +1,22 @@
+// Because Clerk uses short-lived JWTs, it uses middleware that will re-load the JWT from a different page if necessary.
+// This page might return a 401, so we need to tell cypress to ignore the 401 and continue.
```
sdevalk, line 2: Very minor: `Cypress`?

apps/researcher/src/lib/encode-uri-identifier.ts
sdevalk, line 1: - The name, to me, is a bit redundant: `uri-identifier`: a URI is an identifier ('Uniform Resource Identifier'). Rename?
- The name suggests that the file is about 'encoding', but the file also contains a function for 'decoding'.
- Add unit tests for the functions?

apps/researcher/src/lib/encode-uri-identifier.ts
```diff
@@ -0,0 +1,10 @@
+// Linked data uses URIs as identifiers.
+// To use the URIs in dynamic routes, we need to encode them first.
+
+export function encodeUriIdentifier(uri: string) {
+  return encodeURIComponent(uri.replace(/\./g, '%2E'));
+}
+
+export function decodeUriIdentifier(routeSegment: string) {
```
sdevalk, line 8: - `encodeUriIdentifier` uses `uri` as param, this function `routeSegment`. Use a uniform name (e.g. `uri`)?
- This function uses a `replaceAll()` but the `encodeUriIdentifier` a `replace` with a `g` flag. Is it possible to harmonize these?

apps/researcher/src/messages/nl/clerk.ts
```diff
@@ -0,0 +1,659 @@
...
+// With the main difference a rename of 'organisatie' to 'community'
+
+import type {LocalizationResource} from '@clerk/types';
+
+const commonTexts = {
+  signIn: {
+    phoneCode: {
+      title: 'Check je telefoon',
+      subtitle: 'om verder te gaan naar {{applicationName}}',
+      formTitle: 'Verificatie code',
```
sdevalk, line 11: 'Verificatiecode' (one word)

apps/researcher/src/messages/nl/clerk.ts
```diff
@@ -0,0 +1,659 @@
...
+  formFieldLabel__phoneNumber_username: 'telefoonnummer of gebruikersnaam',
+  formFieldLabel__emailAddress_phoneNumber_username:
+    'E-mailadres, telefoonnummer of gebruikersnaam',
+  formFieldLabel__password: 'Wachtwoord',
+  formFieldLabel__newPassword: 'Nieuw wachtwoord',
+  formFieldLabel__confirmPassword: 'Wachtwoord bevestigen',
+  formFieldLabel__firstName: 'Voornaam',
+  formFieldLabel__lastName: 'Achternaam',
+  formFieldLabel__backupCode: 'Backupcode',
+  formFieldLabel__organizationName: 'Organisatienaam',
```
sdevalk, line 38: Or should this be 'communitynaam'?

apps/researcher/src/messages/nl/clerk.ts
```diff
@@ -0,0 +1,659 @@
...
+  formFieldAction__forgotPassword: 'Wachtwoord vergeten?',
+  formFieldHintText__optional: 'Optioneel',
+  formButtonPrimary: 'Doorgaan',
+  signInEnterPasswordTitle: 'Vul je wachtwoord in',
+  backButton: 'Terug',
+  footerActionLink__useAnotherMethod: 'Een andere methode gebruiken',
+  badge__primary: 'Hoofd',
+  badge__thisDevice: 'Dit apparaat',
+  badge__userDevice: 'Gebruikersapparaat',
+  badge__otherImpersonatorDevice: 'Ander immitatie apparaat',
```
sdevalk, line 63: 'immitatieapparaat' (one word)

apps/researcher/src/messages/nl/clerk.ts
```diff
@@ -0,0 +1,659 @@
...
+      blockButton__emailLink: 'Verzend link naar {{identifier}}',
+      blockButton__emailCode: 'Verzend code naar {{identifier}}',
+      blockButton__phoneCode: 'Verzend code naar {{identifier}}',
+      blockButton__password: 'Inloggen met je wachtwoord',
+      blockButton__totp: 'Gebruik je authenticator app',
+      blockButton__backupCode: 'Gebruik een backupcode',
+      getHelp: {
+        title: 'Help',
+        content:
+          'Als je geen toegang hebt neem dan contact op met de klantenservice en we helpen je verder.',
```
sdevalk, line 206: We should probably revise these texts at some point - I don't think we have a 'klantenservice' :)

apps/researcher/src/messages/nl/clerk.ts
```diff
@@ -0,0 +1,659 @@
...
+  },
+  impersonationFab: {
+    title: 'Ingelogd als {{identifier}}',
+    action__signOut: 'Uitloggen',
+  },
+  organizationProfile: {
+    start: {
+      headerTitle__members: 'Leden',
+      headerTitle__settings: 'Instellingen',
+      headerSubtitle__members: 'Organisatieleden bekijken en beheren',
```
sdevalk, line 527: Or 'communityleden'?

apps/researcher/src/messages/nl/clerk.ts
```diff
@@ -0,0 +1,659 @@
...
+      headerSubtitle__settings: 'Community-instellingen beheren',
+    },
+    profilePage: {
+      title: 'Organisatieprofiel',
```
sdevalk, line 531: Or 'communityprofiel'?

apps/researcher/src/middleware.ts
```diff
@@ -9,27 +10,30 @@ export const locales = ['en', 'nl'];
...
-    defaultLocale: 'en',
-  });
+export default authMiddleware({
+  beforeAuth(request: NextRequest) {
+    // This middleware intercepts requests to `/` and will redirect
+    // to one of the configured locales instead (e.g. `/en`).
+    // In the background a cookie is set that will remember the
+    // locale of the last page that the user has visited.
+    // The middleware furthermore passes the resolved locale
+    // to components in your app.
```
sdevalk, line 23: Remove 'in your app'? It sounds a bit strange.

apps/researcher/src/lib/encode-uri-identifier.ts
```diff
@@ -0,0 +1,10 @@
+// Linked data uses URIs as identifiers.
```
sdevalk, line 1: Consideration: remove this line? The frontend is/should be unaware of the use of Linked Data.

sdevalk (commented):
Very cool!

barbarah:
I'm getting errors in the Cypress tests and I cannot find the problem. I'm going to check back tomorrow, hopefully with some fresh ideas.

apps/researcher/src/lib/uri-transformer.ts
```diff
@@ -0,0 +1,8 @@
+// To use the URIs in dynamic routes, we need to encode them first.
+export function encodeUri(uri: string) {
+  return encodeURIComponent(uri.replaceAll('.', '%2E'));
```
sdevalk, line 3: Perhaps add a comment why this replacement is necessary?

apps/researcher/src/lib/uri-transformer.test.ts
```diff
@@ -0,0 +1,26 @@
+import {describe, expect, it} from '@jest/globals';
+import {encodeUri, decodeUri} from './uri-transformer';
+
+describe('encodeUri', () => {
+  it('encodes an url', () => {
+    const encodedUri = encodeUri('https://museum.example.org/');
+    expect(encodedUri).toBe('https%3A%2F%2Fmuseum%252Eexample%252Eorg%2F');
+  });
+
+  it('encodes a string with spaces', () => {
```
sdevalk, line 10: Spaces have their own tests. Perhaps add a comment why?

sdevalk (commented):
Just some minor questions :)

apps/researcher/src/lib/uri-transformer.ts
```diff
@@ -0,0 +1,8 @@
+// To use the URIs in dynamic routes, we need to encode them first.
```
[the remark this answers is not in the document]
barbarah: I don't think it is obvious we use URI in dynamic route segments.

apps/researcher/src/lib/uri-transformer.test.ts
barbarah: No real reason. I just found it hard to think of things to test for one line of code.

apps/researcher/src/lib/uri-transformer.ts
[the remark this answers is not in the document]
sdevalk: 1. I think that would be a comment you could put in the place where these functions are invoked - then the context is clear? The functions in `uri-transformer` are generic, unaware of dynamic route segments.

2. For consideration: is there a library that we can use instead of writing our own encoder and decoder? For example: there are [other characters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent#description) that might need encoding and decoding. Perhaps a third-party lib already takes care of this.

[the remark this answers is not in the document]
barbarah: The error is very weird. Only if the cookie `NEXT_LOCALE` is not set (first-time load or always in Cypress tests) AND the route segment contains a dot. The loading of the translations will fail within the `<ClerkProvider>` container. It will succeed outside of the container.

```
        <NextIntlClientProvider locale={locale} messages={messages}>
          {t('iWillRender')}
          <ClerkProvider localization={clerkLocale}>
             {t('iWillError')}
             {children}
          </ClerkProvider>
        </NextIntlClientProvider>
 ```

The error: "Uncaught Error: Unable to find `next-intl` locale. Have you configured the middleware?"

I'm not sure if the problem lies at `next-intl` or Clerk. But the combination of the two, with certain characters in route segments, is a problem.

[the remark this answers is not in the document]
Comment (barbarah) [reply apps/researcher/src/lib/uri-transformer.ts:1]: I have checked all possible characters. Only the dot is a problem.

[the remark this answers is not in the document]
Comment (barbarah) [reply apps/researcher/src/lib/uri-transformer.ts:1]: I'm in doubt about what I should do now. I think there is a bug in next-intl or Clerk. Replacing the dots is a workaround for this bug.

Using only `encodeURIComponent ` to create valid route segments should be completely valid. So I don't want to introduce a new library.

I think the best for now is to create a bug report. And for now, include the workaround in our code with a comment on why we are replacing dots.

File: apps/researcher/src/lib/route-segment-transformer.test.ts [hunk 14]
```diff
@@ -0,0 +1,21 @@
+import {describe, expect, it} from '@jest/globals';
+import {
+  encodeRouteSegment,
+  decodeRouteSegment,
+} from './route-segment-transformer';
+
+describe('encodeRouteSegment', () => {
+  it('encodes a URI', () => {
```
sdevalk, line 8: Minor: `encodes a route segment`? The segment could be a URI, but doesn't have to be?

apps/researcher/src/lib/uri-transformer.ts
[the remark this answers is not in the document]
sdevalk: Perfect!

I also like the new name, with 'route segments' - it makes it more specific, and doesn't suggest generic functions. If the issue is indeed at Clerk, we could even tailor the file name, e.g. `clerk-route-segment-transformer.ts`, so as to make very clear that the functions are to be used in a specific context.

apps/researcher/src/lib/route-segment-transformer.test.ts
```diff
@@ -0,0 +1,21 @@
+import {describe, expect, it} from '@jest/globals';
+import {
+  encodeRouteSegment,
+  decodeRouteSegment,
+} from './route-segment-transformer';
+
+describe('encodeRouteSegment', () => {
```
sdevalk, line 7: For consideration: the heart of `route-segment-transformer.ts` is about replacing dots. Should we have a test for this, or rename the current test?

apps/researcher/src/lib/route-segment-transformer.test.ts
barbarah: I renamed it to 'encodes a string with dots'

PR #166, opened 2023-07-10
Implement new layout

barbarah:
This is just the first implementation with only the components with existing data. Also, some components still need work (filters, pagination, missing colors). I have discussed this with @Doppen. I need more information about these components and will implement the designs in new tickets.


apps/researcher/src/app/[locale]/error.tsx
```diff
@@ -0,0 +1,15 @@
...
+import {useEffect} from 'react';
+
+export default function Error({error}: {error: Error}) {
+  useEffect(() => {
+    console.error(error);
+  }, [error]);
+
+  return (
+    <div data-testid="error">
+      <h2>Something went wrong!</h2>
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/error.tsx:12]: Is there a way to localize this message? (Also applies to our existing `error.tsx` files)

apps/researcher/src/app/[locale]/objects/[id]/gallery.tsx
```diff
@@ -18,9 +18,10 @@ interface Props {
     src: string;
     alt: string;
   }[];
+  resetText: string;
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/objects/[id]/gallery.tsx:21]: What is a 'reset text', or why and when do you need to reset a text?

apps/researcher/src/app/[locale]/objects/[id]/metadata.tsx
```diff
@@ -0,0 +1,93 @@
...
+  dataKey: '',
+}));
+
+interface Props {
+  dataKey: string;
+  children: ReactNode;
+}
+
+export function MetadataContainer({dataKey, children}: Props) {
+  useCurrentOwner.setState({dataKey: dataKey});
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/objects/[id]/metadata.tsx:18]: Would the shorthand `useCurrentOwner.setState({dataKey});` also work?

apps/researcher/src/app/[locale]/objects/[id]/metadata.tsx
```diff
@@ -0,0 +1,93 @@
...
+import {create} from 'zustand';
+
+export const useCurrentOwner = create(() => ({
+  name: '',
+  location: '',
+  dataKey: '',
+}));
+
+interface Props {
+  dataKey: string;
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/objects/[id]/metadata.tsx:13]: What is a 'data key'? Would it be possible to use a more specific or functional name?

apps/researcher/src/app/[locale]/objects/[id]/metadata.tsx
```diff
@@ -0,0 +1,93 @@
...
+        <div className="w-full xl:w-4/5 flex flex-col gap-2">{children}</div>
+      </div>
+    </div>
+  );
+}
+
+interface MetadataEntryProps {
+  children: ReactNode;
+  isOwner: boolean;
+  isEmpty: boolean;
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/objects/[id]/metadata.tsx:43]: - What does `isEmpty` mean? For example: is there no data?
- Could this component figure out whether there is or isn't data (so that we don't have to use a property for it)?

apps/researcher/src/app/[locale]/objects/[id]/slide-over-gallery.tsx
```diff
@@ -83,18 +108,41 @@ interface OpenSeaDragonViewer {
...
+          },
+        })
+      );
+    };
+
+    initOpenseadragon();
+    return () => {
+      viewer && viewer.destroy();
+    };
+  }, []);
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/objects/[id]/slide-over-gallery.tsx:146]: Is is possible to resolve this warning from GitHub Actions?

`React Hook useEffect has missing dependencies: 'image' and 'viewer'. Either include them or remove the dependency array. If 'setViewer' needs the current value of 'image', you can also switch to useReducer instead of useState and read 'image' in the reducer`

sdevalk (commented):
Nice! I've added some questions.

apps/researcher/src/app/[locale]/objects/[id]/gallery.tsx
Comment (barbarah) [reply apps/researcher/src/app/[locale]/objects/[id]/gallery.tsx]: This is the text in the zoom reset button. I have renamed it to `resetZoomButton` and imported it directly into the gallery component. This is possible since [this pull request](https://github.com/colonial-heritage/colonial-collections/pull/172).

apps/researcher/src/app/[locale]/error.tsx
Comment (barbarah) [reply apps/researcher/src/app/[locale]/error.tsx]: I have removed the old error file. Because this will never load. In the `[locale]` level it is possible to add translations. So I will add it here.

apps/researcher/src/app/[locale]/objects/[id]/metadata.tsx
Comment (barbarah) [reply apps/researcher/src/app/[locale]/objects/[id]/metadata.tsx:43]: Yes, this means there I no data. I will remove it for now and check if the `children` are null.

Eventually, I want a richer layout for the children (see, for example, the tags on materials in the design). Then `children` will not always be null if there is no data, but I will tackle that problem then. Probably by making a separate `<MetadataTextEntry>` component and `<MetadataThesaurusEntry>`. But I'm still discussing the tag-like layout with Bas. That is why this is not yet integrated.

apps/researcher/src/app/[locale]/objects/[id]/slide-over-gallery.tsx
Comment (barbarah) [reply apps/researcher/src/app/[locale]/objects/[id]/slide-over-gallery.tsx:142]: Yes! Sorry, I missed that one.

apps/researcher/src/app/[locale]/objects/[id]/slide-over-gallery.tsx
```diff
@@ -83,18 +104,44 @@ interface OpenSeaDragonViewer {
   image: string;
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/objects/[id]/slide-over-gallery.tsx:104]: Or is this not an image (a representation of something) but a URL? `imageUrl`?

sdevalk (commented):
Nice! Just two questions :)

apps/researcher/src/app/[locale]/objects/[id]/page.tsx
```diff
@@ -86,91 +35,114 @@ export default async function Details({params}: Props) {
...
+                <strong className="text-left text-sky-700 cursor-pointer">
+                  {object.owner?.name}
+                </strong>
+                {/* Add this when location owner information */}
+                {/* <span className="hidden sm:inline"> - </span>
+                <span> Amsterdam</span> */}
+              </span>
+            </span>
           </div>
+          {/* Add when 'object id' is part of API */}
```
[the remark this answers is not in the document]
Comment (barbarah) [reply apps/researcher/src/app/[locale]/objects/[id]/page.tsx:84]: Cool, yes this looks correct.

sdevalk (approved):

PR #172, opened 2023-07-17
Next intl client provider on layout level

barbarah:
In the current code, the client component translations must be passed down from server components through props OR wrapped in a `NextIntlClientProvider`. I decided to place the `NextIntlClientProvider` on the layout level so you can always use translations in client components and never pass down translations through props.

I decided this because passing down props was confusing (see, for example, questions about `resetText` in this [pull request](https://github.com/colonial-heritage/colonial-collections/pull/166)). And in the new version of the [official docs](https://next-intl-docs.vercel.app/docs/getting-started/app-router-server-components) of `next-intl`, using `NextIntlClientProvider` on the layout level is also recommended.


sdevalk (approved):
👍🏻

PR #179, opened 2023-07-20
feat: object fetcher details

sdevalk:
This PR overhauls the frontend API that fetches object information from our backends. The primary change:

The API now only queries Elasticsearch when searching for objects. Previously ES was also queried when fetching the details of an object, but this limits the API's data and functionality to what we have in ES. Instead, the API now queries the SPARQL endpoint when fetching the details of objects, allowing for much richer data.

Scope notes:
1. This PR does not change the API that fetches person information - I'll do this in a separate PR.
2. This PR does not contain provenance information - that's still a part of #152, and requires a rewrite in a separate PR.


barbarah (approved):
Looks good! I always love the integration tests. I read them as documentation. I also like that all definitions are in one file; I often look at those during development. So as a user of this fetcher, I'm very happy.

PR #180, opened 2023-07-20
Layout heritage object card, if there is no image

barbarah:


apps/researcher/src/app/[locale]/(objects)/heritage-object-card.tsx
```diff
@@ -3,44 +3,60 @@ import {useTranslations} from 'next-intl';
...
 interface Props {
   heritageObject: HeritageObject;
 }
 
 export default function HeritageObjectCard({heritageObject}: Props) {
   const t = useTranslations('HeritageObjectCard');
+  const imageUrl =
+    heritageObject.images && heritageObject.images.length > 0
+      ? heritageObject.images[0].contentUrl
+      : null;
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/(objects)/heritage-object-card.tsx:17]: Out of curiosity: when do you use `null` and when `undefined`?

Thinking out load: if `imageUrl` is `null` it suggests that it has a (valid, empty) value. If `imageUrl` is `undefined` it suggests that it has no value.

If I read line 39 correctly, the check for `imageUrl` is not about a specific value but about existence: does an image URL exist?

sdevalk (approved):
Perfect! Just one, non-blocking question :)

PR #191, opened 2023-08-04
feat: organizations

sdevalk:
This PR adds a new endpoint to the frontend API, for fetching the info of a specific organization, e.g. its name, homepage URL and contact details. This info can be used in the [new design](https://gui-prototype.colonialcollections.nl/object_no_community.html), e.g. when a user clicks on 'Contact info'.

The info can be retrieved by creating an instance of `Organizations` (see `src/lib/organizations-instance.ts`) and by calling `.getById(<IRI of the organization>)`.

Beware: this PR builds upon #189. Please first review and merge that one before reviewing and merging this one.


barbarah:
@sdevalk can you merge `main` into this branch or rebase so only the relative commits are visible?

sdevalk:
@barbarah Done!

The e2e test is failing though, with message `Error: API rate limit exceeded for installation ID 33161941.`

barbarah:
Thank you. I will try to rerun the e2e test later.

barbarah:
Maybe you are already aware, but I cannot use this endpoint for production data yet, because `owner` in the `HeritageObject` is always empty. So I do not know the IRI of the organization. In the test data `owner` is filled.

barbarah (approved):
This endpoint contains all information needed for the organization and is in sync with the other endpoints. However, I still need the IRI of the organization (see my last comment) in production data. But this is not part of this pull request.

sdevalk:
That's a good one! We do not have the owners in the real data. We can only present the publishers of the data. I'll create a separate PR that allows you to get the identifier (IRI) of the publisher of a heritage object. You can then use that IRI to fetch the publisher's contact details using the API in this PR.

PR #192, opened 2023-08-06
feat: date created

sdevalk:
This PR adds a `dateCreated` property to a `HeritageObject` and fixes [issue 146](https://github.com/colonial-heritage/sprints/issues/146).


barbarah (approved):
Looks good. I will talk to @Doppen about the design since we have two dates. And the design only shows one.

sdevalk:
Perfect! The two dates are there to express uncertainty: a precise date of creation ("date X") is not always known but a range is ("created between date X and date Y"). This representation of dates might change later, though (e.g. by introducing EDTF in https://github.com/orgs/colonial-heritage/projects/1/views/1?pane=issue&itemId=22566216).

PR #199, opened 2023-08-11
feat: expose dataset publisher

sdevalk:
This PR adds the `publisher` of the dataset to the `HeritageObject`. You can then use the `id` of the publisher to fetch the contact details of the publisher via `Organizations.getById()`.

This resolves [this comment](https://github.com/colonial-heritage/colonial-collections/pull/191#issuecomment-1673330104).

The PR also fixes quite a number of inconsistencies and improves the names of helper functions. These changes shouldn't affect the APIs that the frontend uses.


barbarah (approved):
Nice!

PR #204, opened 2023-08-17
Add basic community pages

barbarah:
closes https://github.com/colonial-heritage/sprints/issues/158

For now, the only way to navigate to the community page is by using the /communities URL. There is no navigation button to this page because I need to figure out where to place the button. (@Doppen?)

The community page can be found here: https://colonial-collections-researcher-git-co-6094dc-colonial-heritage.vercel.app/communities

#  Scope:

## Page 1: An overview of all communities ("/communities")

This page has no design yet, so I will keep this simple.

## Page 2: A community details page. ("/communities/id")

- A list of users
- A title
- A button to join the community, only visible if you have not already joined and logged in
- A button to edit the community, only visible to admins

### Out of scope for this ticket:
- [Adding metadata to Users and Communities](https://github.com/colonial-heritage/sprints/issues/162)
- [Adding the object lists to the page](https://github.com/colonial-heritage/sprints/issues/103)
- Design of the communities page + navigation to this page


apps/researcher/src/app/[locale]/communities/[id]/buttons.tsx
```diff
@@ -0,0 +1,89 @@
...
+        console.error(error);
+      }
+    });
+  };
+
+  const t = useTranslations('Community');
+
+  if (!showJoinButton) {
+    return null;
+  }
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/communities/[id]/buttons.tsx:48]: Perhaps put this on line 27, right after the check on line 26?

apps/researcher/src/app/[locale]/communities/[id]/buttons.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+  const showJoinButton = isLoaded && isSignedIn && !userIsMember;
+
+  const joinCommunityClick = () => {
+    setIsClicked(true);
+    startTransition(async () => {
+      try {
+        await joinCommunity({
+          organizationId: organization.id,
+          userId: user!.id,
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/communities/[id]/buttons.tsx:34]: Minor: this uses `!` but line 22 uses `?` - intentionally?

apps/researcher/src/app/[locale]/communities/[id]/page.tsx
```diff
@@ -0,0 +1,107 @@
...
+  try {
+    organization = await clerkClient.organizations.getOrganization({
+      organizationId: params.id,
+    });
+  } catch (error) {
+    return <div data-testid="no-entity">{t('noEntity')}</div>;
+  }
+
+  const memberships =
+    await clerkClient.organizations.getOrganizationMembershipList({
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/communities/[id]/page.tsx:25]: Could the call to `getOrganizationMembershipList()` fail, and if so, should we catch/handle it here, or can the error bubble up?

apps/researcher/src/app/[locale]/communities/[id]/page.tsx
```diff
@@ -0,0 +1,107 @@
...
+export default async function CommunityPage({params}: Props) {
+  const t = await getTranslations('Community');
+
+  let organization;
+  try {
+    organization = await clerkClient.organizations.getOrganization({
+      organizationId: params.id,
+    });
+  } catch (error) {
+    return <div data-testid="no-entity">{t('noEntity')}</div>;
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/communities/[id]/page.tsx:21]: Is the `error` always about a non-existing organization? Or could it also fail due to other reasons?

apps/researcher/src/app/[locale]/communities/[id]/page.tsx
```diff
@@ -0,0 +1,107 @@
...
+import {JoinCommunityButton, EditCommunityButton} from './buttons';
+
+interface Props {
+  params: {id: string};
+}
+
+export default async function CommunityPage({params}: Props) {
+  const t = await getTranslations('Community');
+
+  let organization;
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/communities/[id]/page.tsx:15]: Thinking out load: would it make sense to move this logic (lines 15 to 27) to a separate file? That way we (1) have less logic in this component, (2) make the logic testable, (3) hide the fact that we're using Clerk?

apps/researcher/src/messages/nl/messages.json
```diff
@@ -129,5 +129,17 @@
...
     "message": "Er is iets fout gegaan!"
+  },
+  "Communities": {
+    "title": "Communities"
+  },
+  "Community": {
+    "title": "Community van",
+    "backButton": "Alle communities",
+    "noEntity": "Community kon niet gevonden worden",
+    "editButton": "community page aanpassen",
```
sdevalk, line 140: Capital `c`: `Community`? :)

apps/researcher/src/app/[locale]/communities/[id]/buttons.tsx
Comment (barbarah) [reply apps/researcher/src/app/[locale]/communities/[id]/buttons.tsx:34]: Yes. I don't want an error if `user` is undefined on row 22. The button will be hidden because `isSignedIn` then is false.

I do want an error if there is no user when trying to save.

But I will change the logic of `showJoinButton` a bit so it will be false if there is no user, which is probably more precise, without making the assumption that `isSignedIn` is false if `user` is undefined.

apps/researcher/src/app/[locale]/communities/[id]/page.tsx
Comment (barbarah) [reply apps/researcher/src/app/[locale]/communities/[id]/page.tsx:21]: It could fail due other reasons, fixed in the refactor.

apps/researcher/src/lib/community.test.ts
```diff
@@ -0,0 +1,161 @@
...
+    const {community, error} = await getCommunity('community1');
+
+    expect(community).toEqual({id: 'community1'});
+    expect(error).toBeUndefined();
+  });
+
+  it('should return an error', async () => {
+    (clerkClient.organizations.getOrganization as jest.Mock).mockImplementation(
+      () => {
+        throw {status: 404};
```
sdevalk, line 34: For clarity (and typing): would it be possible to throw an (instance of) `Error` instead?

apps/researcher/src/lib/community.test.ts
```diff
@@ -0,0 +1,161 @@
...
+  clerkClient: {
+    organizations: {
+      getOrganization: jest.fn(),
+      getOrganizationMembershipList: jest.fn(),
+    },
+  },
+}));
+
+describe('getCommunity', () => {
+  it('should return a community', async () => {
```
sdevalk, line 18: Or active voice: `returns a community`?

apps/researcher/src/lib/community.test.ts
```diff
@@ -0,0 +1,161 @@
...
+    const {memberships, error} = await getMemberships('community1');
+
+    expect(memberships).toEqual([
+      {
+        role: 'role',
+      },
+    ]);
+    expect(error).toBeUndefined();
+  });
+  it('should return an error', async () => {
```
sdevalk, line 64: Minor: add a newline between the tests, for readability?

apps/researcher/src/lib/community.ts
```diff
@@ -0,0 +1,91 @@
...
+
+type Community = {
+  id: string;
+  name: string;
+};
+
+export type Membership = {
+  id: string;
+  role: string;
+  publicUserData?: {
```
sdevalk, line 11: Would it be possible to use a more descriptive name instead? (`publicUserData` sounds a bit vague - what is it?) For example: `user` or `member`?

apps/researcher/src/app/[locale]/communities/page.tsx
```diff
@@ -0,0 +1,21 @@
+import {clerkClient} from '@clerk/nextjs';
+import {getTranslations} from 'next-intl/server';
+
+// For now this is a basic page without design, just for navigating between communities.
+export default async function CommunitiesPage() {
+  const organizations = await clerkClient.organizations.getOrganizationList();
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/communities/page.tsx:6]: Or should we also use `@/lib/community` here (rather than the Clerk client directly)?

apps/researcher/src/app/[locale]/communities/[id]/page.tsx
```diff
@@ -0,0 +1,99 @@
...
+interface Props {
+  params: {id: string};
+}
+
+export default async function CommunityPage({params}: Props) {
+  const t = await getTranslations('Community');
+
+  const {community, error: communityError} = await getCommunity(params.id);
+  const {memberships, error: membershipsError} = await getMemberships(
+    params.id
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/communities/[id]/page.tsx:17]: Would it be possible to use a `try...catch...` instead (to use the language's possibilities to its fullest)? For example:

```
try {
  const community = await getCommunity(params.id);
} catch (err) {
  if (err.status === 404) {
    return <div data-testid="no-entity">{t('noEntity')}</div>;
  }
  return <div data-testid="error">{t('error')}</div>;
}
```

An alternative would be to make `community` `undefined` if no community is found (rather than an error): the absence of a community isn't necessarily an error:

```
let community;
try {
 community = await getCommunity(params.id);
} catch (err) {
  return <div data-testid="error">{t('error')}</div>;
}

if (community === undefined) {
  return <div data-testid="no-entity">{t('noEntity')}</div>;
}
```

apps/researcher/src/lib/community.ts
```diff
@@ -0,0 +1,91 @@
...
+): Promise<CommunityResponse | CommunityErrorResponse> {
+  let community, error;
+
+  try {
+    community = await clerkClient.organizations.getOrganization({
+      organizationId: communityId,
+    });
+  } catch (e) {
+    error = e as Error;
+    return {community: undefined, error};
```
sdevalk, line 44: Would it be an option to just throw the Clerk error (and catch it in the caller's code rather than here)? Or if you don't want Clerk errors to 'leak' into the caller's code: catch it here and throw our own (instance of) `Error`?

apps/researcher/src/lib/community.ts
```diff
@@ -0,0 +1,91 @@
...
+  id: string;
+  name: string;
+};
+
+export type Membership = {
+  id: string;
+  role: string;
+  publicUserData?: {
+    userId: string;
+    firstName: string | null;
```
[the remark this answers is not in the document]
barbarah: This is a subset of the Membership model of Clerk. Clerk uses null instead of undefined. Same for the name `publicUserData`. This is the name Clerk uses.

apps/researcher/src/lib/community.ts
barbarah: I was in doubt which approach to take for error handling. I will just throw the Clerk error.

apps/researcher/src/lib/community.test.ts
barbarah: I tried, but I could not simulate a Clerk error with an instance of an Error. Maybe I needed to do something differently. But for now, I have removed the tests for `getCommunity` and `getMemberships` because I have simplified the functions to only a wrapper around Clerk functions. And then the tests were only testing my Clerk mock.

apps/researcher/src/lib/community.ts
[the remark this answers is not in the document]
barbarah: I have used the Clerk types instead. So it's clear where these types are coming from.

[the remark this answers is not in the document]
sdevalk: My earlier suggestions were about creating a separation between Clerk and our frontend - as a design principle, to prevent tight coupling. If you think that separation is not necessary, I'm all good 😄

apps/researcher/src/app/[locale]/communities/[id]/page.tsx
```diff
@@ -0,0 +1,110 @@
...
+    }
+    return <ErrorMessage error={t('error')} />;
+  }
+
+  try {
+    memberships = await getMemberships(params.id);
+    isAdmin = isAdminOf(memberships);
+  } catch (err) {
+    if ((err as Error).status === 404) {
+      return <ErrorMessage error={t('error')} />;
```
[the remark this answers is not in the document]
Comment (barbarah) [reply apps/researcher/src/app/[locale]/communities/[id]/page.tsx:35]: Sorry, the if statement should have been removed. I always want to show the general error.

apps/researcher/src/lib/community.ts
```diff
@@ -0,0 +1,64 @@
...
+  return clerkClient.organizations.getOrganizationMembershipList({
+    organizationId: communityId,
+  });
+}
+
+export function getAllCommunities(): Promise<Community[]> {
+  return clerkClient.organizations.getOrganizationList();
+}
+
+export function isAdminOf(memberships: ReadonlyArray<Membership>): boolean {
```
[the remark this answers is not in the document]
barbarah: It checks if it's an admin of the memberships group. But I don't have a strong opinion, I will change it.

apps/researcher/src/lib/community.test.ts
```diff
@@ -0,0 +1,94 @@
...
+import {isAdmin, Membership} from './community';
+
+jest.mock('@clerk/nextjs', () => ({
+  auth: jest.fn().mockImplementation(() => ({
+    userId: 'me',
+  })),
+}));
+
+describe('isAdmin', () => {
+  it('returns true is the user is an admin', () => {
```
sdevalk, line 12: Minor: `if the user...`

apps/researcher/src/lib/community.test.ts
```diff
@@ -0,0 +1,94 @@
...
+          firstName: 'Jane',
+          lastName: 'Doe',
+          profileImageUrl: 'https://example.com/image.png',
+        },
+      },
+    ];
+    expect(isAdmin(memberships)).toEqual(true);
+  });
+
+  it('returns false if the user is not an admin', () => {
```
sdevalk, line 38: Minor: omit 'the' from 'the user' (to be in sync with the wordings on line 54)?

apps/researcher/src/app/[locale]/communities/[id]/page.tsx
```diff
@@ -0,0 +1,107 @@
...
+import {getMemberships, getCommunity, isAdmin} from '@/lib/community';
+import ErrorMessage from '@/components/error-message';
+
+interface Props {
+  params: {id: string};
+}
+
+interface Error {
+  status: number;
+}
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/communities/[id]/page.tsx:15]: Sorry, I overlooked this one during my previous review.

You can define your own interface, but I'd say that the error that is returned by `getCommunity()` already has a definition for this? For example, in pseudo-code:

```
import type {SomeClerkError} from 'clerk';

try {
  community = await getCommunity(params.id);
} catch (err) {
  const error = err as SomeClerkError;
  if (error.status === 404) {
    ...
}
```

apps/researcher/src/lib/community.test.ts
```diff
@@ -0,0 +1,94 @@
...
+          firstName: 'John',
+          lastName: 'Doe',
+          profileImageUrl: 'https://example.com/image.png',
+        },
+      },
+    ];
+    expect(isAdmin(memberships)).toEqual(false);
+  });
+
+  it('returns false is user is not an member', () => {
```
sdevalk, line 54: Minor: `a member`?

apps/researcher/src/lib/community.test.ts
```diff
@@ -0,0 +1,94 @@
...
+          firstName: 'John',
+          lastName: 'Doe',
+          profileImageUrl: 'https://example.com/image.png',
+        },
+      },
+    ];
+    expect(isAdmin(memberships)).toEqual(false);
+  });
+
+  it('returns false if memberships is empty', () => {
```
sdevalk, line 90: Or, alternatively: `returns false if there are no memberships` (more functional wordings, less technical)?

apps/researcher/src/app/[locale]/communities/[id]/page.tsx
Comment (sdevalk) [reply apps/researcher/src/app/[locale]/communities/[id]/page.tsx:15]: Ah, too bad (and really strange -- they do not export their error definitions?)!

Two considerations:

1. Are we allowed to import `@clerk/shared` (e.g. in our `devDependencies`)? Or is package for internal use by Clerk only?
2. If (1) is not an option: perhaps we can rename the current `interface Error` to something descriptive, so as to prevent confusion with JavaScript's default `Error` and to make clear that we're defining an interface that should have come from Clerk? For example: `interface ClerkError` (+ a comment explaining the reason why this interface exists)?

Comment (barbarah) [reply apps/researcher/src/app/[locale]/communities/[id]/page.tsx:15]: Solved by installing @clerk/shared as a dev dependency.

sdevalk (approved):
Nice! Just one minor comment :)

PR #211, opened 2023-08-24
fix: made 'dataset' props optional

sdevalk:
The properties of a `Dataset` (e.g. its name and publisher) do not always exist in the data that data providers give us. Our code, however, assumes these properties are there. This PR makes the code a bit more liberal: if the props exist, we use them, otherwise we continue with the data that we have.

This is a small breaking change: [the `isPartOf` prop](https://github.com/colonial-heritage/colonial-collections/blob/fix-objects-dataset-optional/apps/researcher/src/lib/api/definitions.ts#L59) has now become optional.


barbarah (approved):
Nice fix.

PR #224, opened 2023-09-06
Show organization on details page

barbarah:
image

### DIFF
--- a/apps/researcher/.env.development
+++ b/apps/researcher/.env.development
@@ -5,3 +5,5 @@
 SEARCH_PLATFORM_ELASTIC_ENDPOINT_URL=https://api.colonial-heritage.triply.cc/datasets/testing/researcher-search-v1/services/researcher-search-v1-es/elasticsearch
 SEARCH_PLATFORM_SPARQL_ENDPOINT_URL=https://api.colonial-heritage.triply.cc/datasets/testing/researcher-search-v1/services/researcher-search-v1-sparql/sparql
 KG_SPARQL_ENDPOINT_URL=https://api.colonial-heritage.triply.cc/datasets/testing/researcher-entities-v1/services/researcher-entities-v1-sparql/sparql
+
+DATASET_BROWSER_URL="https://dev.datasets.colonialcollections.nl"
--- a/apps/researcher/.env.development
+++ b/apps/researcher/.env.development
@@ -6,4 +6,4 @@ SEARCH_PLATFORM_ELASTIC_ENDPOINT_URL=https://api.colonial-heritage.triply.cc/dat
 SEARCH_PLATFORM_SPARQL_ENDPOINT_URL=https://api.colonial-heritage.triply.cc/datasets/testing/researcher-search-v1/services/researcher-search-v1-sparql/sparql
 KG_SPARQL_ENDPOINT_URL=https://api.colonial-heritage.triply.cc/datasets/testing/researcher-entities-v1/services/researcher-entities-v1-sparql/sparql
 
-DATASET_BROWSER_URL="https://dev.datasets.colonialcollections.nl"
+DATASET_BROWSER_URL=https://dev.datasets.colonialcollections.nl
--- a/apps/researcher/src/app/[locale]/objects/[id]/metadata.tsx
+++ b/apps/researcher/src/app/[locale]/objects/[id]/metadata.tsx
@@ -37,15 +37,15 @@ export function MetadataContainer({identifier, children}: Props) {
 
 interface MetadataEntryProps {
   children: ReactNode;
-  isCurrentLocation: boolean;
+  isCurrentPublisher: boolean;
 }
 
 export function MetadataEntry({
   children,
-  isCurrentLocation = false,
+  isCurrentPublisher = false,
 }: MetadataEntryProps) {
   const {identifier} = useMetadata.getState();
-  const organization = useCurrentPublisher.getState();
+  const {organization} = useCurrentPublisher.getState();
   const t = useTranslations('ObjectDetails');
 
   if (!children) {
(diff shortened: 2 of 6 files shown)


apps/researcher/src/app/[locale]/objects/[id]/metadata.tsx
```diff
@@ -39,48 +37,48 @@ export function MetadataContainer({identifier, children}: Props) {
 
 interface MetadataEntryProps {
   children: ReactNode;
-  isOwner: boolean;
+  isCurrentLocation: boolean;
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/objects/[id]/metadata.tsx:40]: Is the 'location' the place where the heritage object is being held? If so, this is not in the data that we currently have. But we do know the publisher of the data about the heritage object. Replace with e.g. 'isCurrentPublisher'?

apps/researcher/src/app/[locale]/objects/[id]/metadata.tsx
```diff
@@ -39,48 +37,48 @@ export function MetadataContainer({identifier, children}: Props) {
...
-          </div>
-          {isOwner && <div>{t('currentOwner')}</div>}
-        </div>
-        <div className="flex shrink">
-          <div>
-            <button className="flex items-center py-2 px-3 rounded-full bg-sand-100 text-sand-900 hover:bg-white transition text-xs">
-              <HandThumbUpIcon className="w-4 h-4 stroke-sand-600" />
-              &nbsp;(0)
-            </button>
+            {t.rich('addedBy', {
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/objects/[id]/metadata.tsx:67]: Minor - not something to resolve now, but to think about: the data doesn't say that the data publisher has 'added' this information - it could have been added by another organization. But we do now that the information has been 'provided' by the publisher.

apps/researcher/src/app/[locale]/objects/[id]/page.tsx
```diff
@@ -15,6 +24,40 @@ interface Props {
...
+            <br /> {organization.address?.postalCode}{' '}
+            {organization.address?.addressLocality},
+            <br /> {organization.address?.addressCountry}
+          </span>
+          {organization.url && (
+            <a href={organization.url}>{t('linkToWebsite')}</a>
+          )}{' '}
+          {datasetId && (
+            <a
+              href={`https://datasets.colonialcollections.nl/nl/datasets/${encodeURIComponent(
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/objects/[id]/page.tsx:48]: Is there a way to not hard-code this, and to fetch the base URL ('https://datasets.colonialcollections.nl') from a config? Then we can also link to the dev server ('https://dev.datasets.colonialcollections.nl/').

apps/researcher/src/app/[locale]/objects/[id]/page.tsx
```diff
@@ -25,7 +68,11 @@ export default async function Details({params}: Props) {
     return <div data-testid="no-entity">{t('noEntity')}</div>;
   }
 
-  useCurrentOwner.setState({name: object.owner?.name});
+  let organization = null;
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/objects/[id]/page.tsx:71]: A recurring question :) Is this a `null` or an `undefined`? `organizations.getById()` returns the latter.

apps/researcher/src/app/[locale]/objects/[id]/useCurrentPublisher.ts
```diff
@@ -0,0 +1,16 @@
+import {create} from 'zustand';
+import {Organization} from '@/lib/api/definitions';
+
+export default create<Organization>(() => ({
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/objects/[id]/useCurrentPublisher.ts:4]: Thinking out load: is there a way to not create a copy of the 'Organization' data structure (which can change), but to re-use the existing one, with new defaults (empty strings)?

apps/researcher/src/app/[locale]/objects/[id]/metadata.tsx
Comment (barbarah) [reply apps/researcher/src/app/[locale]/objects/[id]/metadata.tsx:67]: Good point. Maybe also something to think about in the design. Now the text in the design is:

image

apps/researcher/src/app/[locale]/objects/[id]/page.tsx
Comment (barbarah) [reply apps/researcher/src/app/[locale]/objects/[id]/page.tsx:71]: This should be undefined.

apps/researcher/src/app/[locale]/objects/[id]/page.tsx
Comment (barbarah) [reply apps/researcher/src/app/[locale]/objects/[id]/page.tsx]: I have moved the URL to environment variables and added the variable to Vercel.

apps/researcher/.env.development
```diff
@@ -5,3 +5,5 @@
 SEARCH_PLATFORM_ELASTIC_ENDPOINT_URL=https://api.colonial-heritage.triply.cc/datasets/testing/researcher-search-v1/services/researcher-search-v1-es/elasticsearch
 SEARCH_PLATFORM_SPARQL_ENDPOINT_URL=https://api.colonial-heritage.triply.cc/datasets/testing/researcher-search-v1/services/researcher-search-v1-sparql/sparql
 KG_SPARQL_ENDPOINT_URL=https://api.colonial-heritage.triply.cc/datasets/testing/researcher-entities-v1/services/researcher-entities-v1-sparql/sparql
+
+DATASET_BROWSER_URL="https://dev.datasets.colonialcollections.nl"
```
sdevalk, line 9: For consistency with the other vars: remove the quotes and the newline on line 8?

apps/researcher/src/app/[locale]/objects/[id]/page.tsx
```diff
@@ -15,6 +24,53 @@ interface Props {
...
+            {organization.address?.addressLocality},
+            <br /> {organization.address?.addressCountry}
+          </span>
+          {organization.url && (
+            <a href={organization.url}>{t('linkToWebsite')}</a>
+          )}{' '}
+          {datasetId && (
+            <a
+              href={`${
+                process.env['DATASET_BROWSER_URL']
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/objects/[id]/page.tsx:62]: Would it be possible to first `import {env} from 'node:process'` and then to use `env['DATASET_BROWSER_URL']` (see e.g. https://2ality.com/2021/12/node-protocol-imports.html)?

sdevalk (approved):
Nice! Just two small questions :)

PR #231, opened 2023-09-11
Add date made to object details

barbarah:
For now, we decided to show the date made as a range with full dates (not just years). There is a big change this will change in the future.


apps/researcher/src/app/[locale]/objects/[id]/date-created-to-string.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+import {describe, expect, it} from '@jest/globals';
+
+describe('dateCreatedToString', () => {
+  it('returns undefined if both startDate and endDate are undefined', async () => {
+    const date = await dateCreatedToString(
+      {startDate: undefined, endDate: undefined, id: 'timeSpan'},
+      'en'
+    );
+
+    expect(date).toBeUndefined();
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/objects/[id]/date-created-to-string.test.tsx:11]: For consistency, use the same `expect()` notation as in the other tests (`expect(dateCreatedToString(...))`)?

Personally I prefer the notation you're using in this test: first the function call, then the expect (I like the readability).

apps/researcher/src/app/[locale]/objects/[id]/date-created-to-string.ts
```diff
@@ -0,0 +1,24 @@
+import {TimeSpan} from '@/lib/api/definitions';
+
+export default function dateCreatedToString(
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/objects/[id]/date-created-to-string.ts:3]: Would it be an option to not use `default`? That way we can add new functions to the file.

apps/researcher/src/app/[locale]/objects/[id]/date-created-to-string.test.tsx
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/objects/[id]/date-created-to-string.test.tsx:1]: Change the file's extension to `.ts`?

apps/researcher/src/app/[locale]/objects/[id]/date-created-to-string.test.tsx
```diff
@@ -0,0 +1,57 @@
+import dateCreatedToString from './date-created-to-string';
+import {describe, expect, it} from '@jest/globals';
+
+describe('dateCreatedToString', () => {
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/objects/[id]/date-created-to-string.test.tsx:4]: Perhaps also add a test with a startDate and/or endDate that use a full date notation (i.e. including time)? Then we can test whether `dateStyle: 'medium',` works.

apps/researcher/src/messages/en/messages.json
```diff
@@ -79,7 +77,9 @@
     "creatorsSubTitle": "",
     "inscriptionsSubTitle": "",
     "addedBy": "Added by",
-    "currentOwner": "Current owner"
+    "currentOwner": "Current owner",
+    "dateCreated": "Date Made",
+    "dateCreatedSubTitle": "When was the object made. Could be a exact or a date range."
```
sdevalk, line 82: 1. `made?`
1. `an exact date or a range`?

apps/researcher/src/app/[locale]/objects/[id]/date-created-to-string.ts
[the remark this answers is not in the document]
Comment (barbarah) [reply apps/researcher/src/app/[locale]/objects/[id]/date-created-to-string.ts:3]: I deliberately chose to make this function specific for `dateCreated` because there has been some discussion about how to display dateCreated. For now, it is a pretty straightforward time span, but I think this is going to change in the future.

Comment (barbarah) [reply apps/researcher/src/app/[locale]/objects/[id]/date-created-to-string.ts:3]: I prefer `export default` for single-function files. But this could be because this rule was part of the Airbnb eslint config I'm more used to working with. Choosing `export default` for single-function files is a more straightforward rule. 

I have no problem changing this, but then, for me, I'm not sure when you prefer `export default` and when `export`. Can you explain your preference some more? For example, do you prefer' export' for single component files? I was not planning to add extra functions to this file. So for me, `export default` makes more sense.

apps/researcher/src/app/[locale]/objects/[id]/date-created-to-string.test.tsx
Comment (barbarah) [reply apps/researcher/src/app/[locale]/objects/[id]/date-created-to-string.test.tsx:4]: Good point!

apps/researcher/src/app/[locale]/objects/[id]/date-created-to-string.ts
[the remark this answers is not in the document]
Comment (sdevalk) [reply apps/researcher/src/app/[locale]/objects/[id]/date-created-to-string.ts:3]: That's fine! I'd still recommend changing the suffix of the function name `dateCreatedToString` though, to make the meaning of the transformation clear (from a date to a formatted date/localized date).

Comment (sdevalk) [reply apps/researcher/src/app/[locale]/objects/[id]/date-created-to-string.ts:3]: I prefer to not use `export default` because of its side-effects (see e.g. https://www.lloydatkinson.net/posts/2022/default-exports-in-javascript-modules-are-terrible/), unless a framework wants it so (e.g. React components).

apps/researcher/src/app/[locale]/objects/[id]/date-created-to-string.test.ts
```diff
@@ -0,0 +1,70 @@
+import dateCreatedToString from './date-created-to-string';
+import {describe, expect, it} from '@jest/globals';
+
+describe('dateCreatedToString', () => {
+  it('returns undefined if both startDate and endDate are undefined', () => {
+    const date = dateCreatedToString(
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/objects/[id]/date-created-to-string.test.ts:6]: Minor: could `date` be renamed to make its meaning clear? The name `date` could suggest it's a `Date`, but it's actually a formatted/localized date.

apps/researcher/src/app/[locale]/objects/[id]/date-created-to-string.test.ts
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/objects/[id]/date-created-to-string.test.ts:1]: The current tests only use `en` as locale. Is it helpful to also test for a different locale (e.g. `nl`)?

apps/researcher/src/app/[locale]/objects/[id]/date-created-formatter.ts
```diff
@@ -0,0 +1,21 @@
+import {TimeSpan} from '@/lib/api/definitions';
+
+export function dateCreatedFormatter(dateCreated: TimeSpan, locale: string) {
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/objects/[id]/date-created-formatter.ts:3]: Nice solution :) Just one detail: in the code base we do use names ending with `..ter` (e.g. `LabelFetcher`), but as far as I can see not for functions. Use a verb instead, for consistency - e.g. `formatDateCreated()`?

sdevalk (approved):

PR #238, opened 2023-09-14
Layout communities page with searching, sorting and pagination

barbarah:
Based on this design

image

# What have I changed:

- There is a navigation button to the communities page in the header
- If you click on a community, you go to `/communities/[slug]` instead of `/communities/[id]`. Since Clerk already provides a slug.
- I have used the `list-store`. But I needed some changes because the SortBy emun is different in this list.

# Out of scope

- I still need to add properties that aren't available. This is the community description and list count.  
- The list cannot be sorted on membership count. This is because of a bug in Clerk. See discord: https://discord.com/channels/856971667393609759/1151078450627624970


apps/researcher/src/lib/community.ts
```diff
@@ -28,8 +31,58 @@ export async function getMemberships(
...
+}
+
+interface GetAllCommunitiesProps {
+  query?: string;
+  sortBy?: CommunitySortBy;
+  limit?: number;
+  offset?: number;
+}
+
+export async function getAllCommunities({
```
sdevalk, line 68: Minor: could we rename the function to `getCommunities()` (which implies 'all')? Or is there a difference between getting 'communities' and 'all communities'?

apps/researcher/src/lib/community.test.ts
```diff
@@ -92,3 +97,115 @@ describe('isAdmin', () => {
...
+      createdAt: 1650000000000,
+    },
+    {
+      id: 'community5',
+      name: 'Community 5',
+      slug: 'community-5',
+      imageUrl: 'https://example.com/image.png',
+      createdAt: 1670000000000,
+    },
+  ];
```
sdevalk, line 138: Add a newline between 138 and 139, for readability?

apps/researcher/src/lib/community.test.ts
```diff
@@ -92,3 +97,115 @@ describe('isAdmin', () => {
...
+  it('sorts communities by name ascending', () => {
```
sdevalk, line 139: Perhaps `sorts communities by name in ascending order` (to make it read like a natural sentence)?

apps/researcher/src/lib/community.test.ts
```diff
@@ -92,3 +97,115 @@ describe('isAdmin', () => {
...
+    const sortedCommunities = sortCommunities(
+      communities,
+      CommunitySortBy.NameAsc
+    );
+    const sortedCommunitiesIds = sortedCommunities.map(
+      community => community.id
+    );
+    expect(sortedCommunitiesIds).toEqual([
```
sdevalk, line 147: Curious: does `toEqual()` test the order of the elements in the array? Or should we use `toStrictEqual()` instead?

apps/researcher/src/lib/community.test.ts
```diff
@@ -92,3 +97,115 @@ describe('isAdmin', () => {
...
+    },
+  ];
+  it('sorts communities by name ascending', () => {
+    const sortedCommunities = sortCommunities(
+      communities,
+      CommunitySortBy.NameAsc
+    );
+    const sortedCommunitiesIds = sortedCommunities.map(
+      community => community.id
+    );
```
sdevalk, line 146: Minor: other tests have a newline before the `expect()` - also do it here?

sdevalk, line 146: Is this `map` required for the test? If I understand the test correctly you want to validate whether `sortedCommunities` have been sorted. But the `expect()` does not use that variable but a new one, `sortedCommunitiesIds`?

apps/researcher/src/lib/community.test.ts
```diff
@@ -92,3 +97,115 @@ describe('isAdmin', () => {
...
+    expect(sortedCommunitiesIds).toEqual([
+      'community2',
+      'community4',
+      'community5',
+      'community3',
+      'community1',
+    ]);
+  });
+
+  it('works with an empty array', () => {
```
sdevalk, line 190: 1. Instead of: `array`: `list` (to use the same words as on e.g. line 195)?
2. Thinking out load: what is the purpose of this test? 'works with' is a bit vague - what does it mean? For example: do you want to test whether there are no side-effects when passing an empty list (e.g. to test that the function doesn't add data)?

apps/researcher/src/lib/community.ts
```diff
@@ -28,8 +31,58 @@ export async function getMemberships(
   });
 }
 
-export function getAllCommunities(): Promise<Community[]> {
-  return clerkClient.organizations.getOrganizationList();
+export enum CommunitySortBy {
```
sdevalk, line 34: Perhaps rename `CommunitySortBy` to `SortBy` (since the file (`communities`) already makes the scope clear)?

apps/researcher/src/lib/community.ts
```diff
@@ -28,8 +31,58 @@ export async function getMemberships(
...
+  NameAsc = 'nameAsc',
+  NameDesc = 'nameDesc',
+  CreatedAtDesc = 'createdAtDesc',
+  MembershipCountDesc = 'membershipCountDesc',
+}
+
+export const defaultCommunitySortBy: CommunitySortBy =
```
sdevalk, line 41: Curious: is the explicit type `CommunitySortBy` required by TypeScript? Or can we omit it?

apps/researcher/src/lib/community.ts
```diff
@@ -28,8 +31,58 @@ export async function getMemberships(
...
+  CommunitySortBy.CreatedAtDesc;
+
+export function sortCommunities(
```
sdevalk, line 44: Or just `sort` (since the file already makes the scope clear - it's about communities)?

apps/researcher/src/lib/community.ts
```diff
@@ -28,8 +31,58 @@ export async function getMemberships(
...
+  communities: Community[],
+  sortBy: CommunitySortBy
+) {
+  return [...communities].sort((a, b) => {
```
sdevalk, line 48: Perhaps also add a comment here, explaining why we're not sorting by `MembershipCountDesc`?

packages/list-store/src/client-list-store.tsx
```diff
@@ -2,17 +2,18 @@
 
 import {useEffect} from 'react';
 import {useListStore} from './useListStore';
-import {SortBy} from './sort';
 import {useSearchParamsUpdate} from './useSearchParamsUpdate';
+import {defaultSortBy as packageDefaultSortBy} from './sort';
```
sdevalk, line 6: The name `packageDefaultSortBy` is a bit unclear to me - what does it mean? Is there a more descriptive name?

packages/list-store/src/search-params.test.ts
```diff
@@ -45,14 +45,15 @@ const sortOptions = {
...
-    expect(getUrlWithSearchParams({})).toBe('/');
+  it('returns "/" if there are no filter options', () => {
+    expect(getUrlWithSearchParams({defaultSortBy: defaultSortBy})).toBe('/');
   });
 
   it('returns "/" with only default values', () => {
     const options = {
       offset: 0,
       sortBy: defaultSortBy,
+      defaultSortBy: defaultSortBy,
```
sdevalk, line 56: Or just `defaultSortBy,`?

apps/researcher/src/app/[locale]/communities/page.tsx
```diff
@@ -1,28 +1,90 @@
...
+  };
+  searchParams?: {
+    query?: string;
+    sortBy?: CommunitySortBy;
+    offset?: number;
+  };
+}
+
+// Revalidate the page
+export const revalidate = 0;
```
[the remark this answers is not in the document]
Comment (barbarah) [reply apps/researcher/src/app/[locale]/communities/page.tsx:25]: I didn't want to add the whole expected sortedCommunities to each test. But maybe that is more clear, so I will just add the whole array.

apps/researcher/src/lib/community.test.ts
barbarah: I didn't want to add the whole expected sortedCommunities to each test. But maybe that is more clear, so I will just add the whole array.

apps/researcher/src/lib/community.test.ts
barbarah: It does test the order, but I will change it to `toStrictEqual()`

apps/researcher/src/lib/community.test.ts
barbarah: I think I'm testing that it won't error. However, I did not have a well-thought-out reason for this test. So I will remove it.

packages/list-store/src/client-list-store.tsx
barbarah: It's the default of the defaultSortBy :)

But I will make defaultSortBy non-optional.

apps/researcher/src/app/[locale]/communities/page.tsx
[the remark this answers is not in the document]
Comment (barbarah) [reply apps/researcher/src/app/[locale]/communities/page.tsx:25]: It may take a while before new communities are visible. But I will remove this for now and see if this is workable.

apps/researcher/src/app/[locale]/communities/page.tsx
```diff
@@ -1,28 +1,79 @@
-import {getAllCommunities} from '@/lib/community';
-import {getTranslations} from 'next-intl/server';
+import {getCommunities, SortBy, defaultSortBy} from '@/lib/community';
+import {getTranslator} from 'next-intl/server';
 import ErrorMessage from '@/components/error-message';
+import CommunityCard from './community-card';
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/communities/page.tsx:5]: Remove the newline?

sdevalk (approved):
Nice! Just one small comment.

PR #244, opened 2023-09-19
dataset browser maintenance

sdevalk:
This is a (non-breaking) maintenance PR for the Dataset Browser — it does not introduce new features:

1. Kinsuk and I have created two new triplestores, a ‘knowledge graph’ and a ‘search graph’. These triplestores replace the temporary triplestores that we’ve used so far. The IRIs of the new triplestores for testing are in the `*.env` files.
2. I’ve temporarily removed the facet ‘spatialCoverage’. So far we’ve used geographical locations from the NMVW thesaurus for this facet, but this source is going to be replaced by a new one, GeoNames.
3. The facets no longer use IRIs for identifying e.g. ‘licenses’ or ‘genres’. That’s because Elasticsearch, a full-text search engine, doesn’t play well with IRIs, only with names/labels. This approach has already been implemented in the Research App some time ago, but now also in the Dataset Browser.

I’ll soon create another PR for updating the Research App, so that it can also make use of the new triplestores.

### DIFF
--- a/apps/dataset-browser/.env.development
+++ b/apps/dataset-browser/.env.development
@@ -2,5 +2,5 @@
 # DO NOT ADD SECRETS TO THIS FILE
 # If you want to add secrets use `.env.development.local` instead
 
-SEARCH_PLATFORM_ELASTIC_ENDPOINT_URL=https://api.colonial-heritage.triply.cc/datasets/testing/dataset-search-v2/services/dataset-search-v2-es/elasticsearch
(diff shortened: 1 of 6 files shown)


sdevalk:
Warning: the Vercel staging environment currently still uses the (IRIs of the) old triplestores, so this PR could fail there. I'll update the IRIs when this PR is approved.

barbarah (approved):
The end-to-end tests fail because the Vercel staging environment currently still uses the (IRIs of the) old triplestore. I checked locally with the new triplestore, then the tests were successful.

PR #246, opened 2023-09-22
Revalidate '/communities' after new community is created

barbarah:
In the current situation, a new community will not appear in the list after creating a community. This is because the page is cached.

To fix this problem, I created the `/revalidate` page. On this page, the list will be revalidated and then redirected.

This must be done on a page, not a callback function, because Clerk only gives an `afterCreateOrganizationUrl` option. I have seen some discussion in Discord to add a callback, but it is unavailable now.

I made this into a general `/revalidate` page so I can use the same logic after deleting an organization. But then I realized that there is no `afterDestroyOrganizationUrl`. I will ask this on the Clerk Discord.

For now, I have added `revalidatePath('/[locale]/communities')` on places that could indicate a deleted organization. So if a user deletes an organization, it will show up in the list, but after a refresh, it is gone. If this is not good enough, we must re-add `export const revalidate = 0;` to the list. But this scenario will only happen sometimes, and it would be a shame if we don't use the cache for a list that will only change a little.

If somehow this revalidate logic didn't trigger, I have added a `revalidate` with a value of one day.

### DIFF
--- a/apps/researcher/src/app/[locale]/communities/community-card.tsx
+++ b/apps/researcher/src/app/[locale]/communities/community-card.tsx
@@ -20,7 +20,7 @@ async function MembershipCount({communityId, locale}: MembershipCountProps) {
   } catch (error) {
     // This could be a sign of a deleted community in the cache.
     // So, revalidate the communities page.
-    revalidatePath('/[locale]/communities');
+    revalidatePath('/[locale]/communities', 'page');
   }
 
   return t.rich('membershipCount', {
--- a/apps/researcher/src/app/[locale]/communities/community-card.tsx
+++ b/apps/researcher/src/app/[locale]/communities/community-card.tsx
@@ -18,9 +18,9 @@ async function MembershipCount({communityId, locale}: MembershipCountProps) {
   try {
     memberships = await getMemberships(communityId);
   } catch (error) {
-    // This could be a sign that the community has been deleted.
-    // In that case, we should revalidate the communities page.
-    revalidatePath('/communities');
+    // This could be a sign of a deleted community in the cache.
+    // So, revalidate the communities page.
+    revalidatePath('/[locale]/communities');
   }
 
   return t.rich('membershipCount', {
--- a/apps/researcher/src/app/[locale]/communities/community-card.tsx
+++ b/apps/researcher/src/app/[locale]/communities/community-card.tsx
@@ -5,6 +5,7 @@ import Link from 'next-intl/link';
 import Image from 'next/image';
 import {Suspense} from 'react';
 import {revalidatePath} from 'next/cache';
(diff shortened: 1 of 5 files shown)


apps/researcher/src/app/[locale]/revalidate/page.tsx
```diff
@@ -0,0 +1,21 @@
...
+import {revalidatePath} from 'next/cache';
+
+interface Props {
+  searchParams?: {
+    path?: string;
+    redirect?: string;
+  };
+}
+
+export default async function RevalidatePaths({searchParams = {}}: Props) {
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/revalidate/page.tsx:11]: To be sure: the interface definition of `Props` says `searchParams` can be undefined. But in the function signature `{}` is always set to `searchParams`. Is the interface definition correct then - shouldn't the `?` be omitted?

Comment (sdevalk) [inline apps/researcher/src/app/[locale]/revalidate/page.tsx:11]: Minor: is this indeed an async function? There are no awaits?

apps/researcher/src/app/[locale]/communities/community-card.tsx
```diff
@@ -12,7 +13,15 @@ interface MembershipCountProps {
 
 async function MembershipCount({communityId, locale}: MembershipCountProps) {
   const t = await getTranslator(locale, 'Communities');
-  const memberships = await getMemberships(communityId);
+
+  let memberships = [];
+  try {
+    memberships = await getMemberships(communityId);
+  } catch (error) {
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/communities/community-card.tsx:20]: Use `err` instead, to be in line with line 24 in file `apps/researcher/src/app/[locale]/communities/[slug]/page.tsx`?

apps/researcher/src/app/[locale]/communities/community-card.tsx
```diff
@@ -12,7 +13,15 @@ interface MembershipCountProps {
...
+    // This could be a sign of a deleted community in the cache.
+    // So, revalidate the communities page.
+    revalidatePath('/[locale]/communities', 'page');
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/communities/community-card.tsx:23]: Is it useful to check the `error` for a specific 'the memberships have been deleted' error (e.g. HTTP status 404 or 410) before revalidating the path?

apps/researcher/src/app/[locale]/revalidate/page.tsx
Comment (barbarah) [reply apps/researcher/src/app/[locale]/revalidate/page.tsx:11]: I think it's best to make `searchParams` mandatory.

apps/researcher/src/app/[locale]/communities/community-card.tsx
Comment (barbarah) [reply apps/researcher/src/app/[locale]/communities/community-card.tsx:23]: Yes. During testing the response was always `404`. Just to be sure I will also check for `410`

sdevalk (approved):
Nice!

PR #248, opened 2023-09-25
Object lists in community page

barbarah:
This is the updated community page with object lists and creating new object lists based on this design: https://gui-prototype.colonialcollections.nl/community.html

# Out of scope

- Community description
- Adding objects to the object list
- List page


packages/database/src/db/schema.ts
```diff
@@ -64,4 +64,9 @@ export const objectItemsRelations = relations(objectItems, ({one}) => ({
   }),
 }));
 
+export const insertObjectListSchema = createInsertSchema(objectLists, {
+  // Don't allow empty strings
```
sdevalk, line 68: Is the comment necessary (or does the rule speak for itself)?

packages/database/src/object-list.ts
```diff
@@ -10,33 +8,43 @@ async function getListsByCommunityId(communityId: string) {
   });
 }
 
-async function getCommunityListsWithObjects(communityId: string) {
+interface GetCommunityListsWithObjectsProps {
+  communityId: string;
+  limitObjects?: number;
+}
+
+async function getCommunityListsWithObjects({
```
sdevalk, line 16: Sorry, I missed this one during a previous review: would it be possible to directly export the functions, without using an `export const objectList`? This way it is immediately clear which function can be used externally and which one cannot.

barbarah: I did this so I can import `objectList` and do, for example `objectList.getCommunityListsWithObjects()`. So, all the functions are grouped into one object. I thought grouping the individual functions would be nice if the database package grew.

apps/researcher/src/app/[locale]/communities/community-card.tsx
```diff
@@ -32,6 +33,26 @@ async function MembershipCount({communityId, locale}: MembershipCountProps) {
...
+}
+
+async function ObjectListCount({communityId, locale}: MembershipCountProps) {
+  const t = await getTranslator(locale, 'Communities');
+
+  let objectLists = [];
+  try {
+    objectLists = await objectList.getListsByCommunityId(communityId);
+  } catch (err) {
+    // Don't error on the overview page. Catch this error on the detail page.
```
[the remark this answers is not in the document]
Comment (sdevalk) [reply apps/researcher/src/app/[locale]/communities/community-card.tsx]: Or perhaps console.error(err) the error and update the comment, to make the intention clearer? It's now a bit confusing to me, with a reference to an overview page and a detail page.

packages/database/src/object-list.ts
sdevalk: I understand. I find the names a bit hard to follow, though - e.g. `objectList.getCommunityListsWithObjects()` or `objectList.addObjectToList()` (lots of 'list' and 'object' :)).

If grouping of functions is important, would it be an idea to use a class instead? For example (pseudo code):

```
export class ObjectList {
  getByCommunityId(id, {withObjects: true}) // was: getListsByCommunityId and getCommunityListsWithObjects
  create() // was: createListForCommunity
  addObject() // was: addObjectToList
}
```

apps/researcher/src/lib/community.ts
```diff
@@ -23,6 +23,12 @@ export async function getCommunityBySlug(slug: string): Promise<Community> {
   });
 }
 
+export async function getCommunityById(id: string): Promise<Community> {
```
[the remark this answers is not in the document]
barbarah: Yes, the actual response still contains all properties. I have removed the subtype `Community` because this is confusing.

apps/researcher/src/app/[locale]/communities/community-card.tsx
```diff
@@ -32,6 +34,27 @@ async function MembershipCount({communityId, locale}: MembershipCountProps) {
...
+  let objectLists = [];
+  try {
+    objectLists = await objectList.getByCommunityId(communityId);
+  } catch (err) {
+    console.error(err);
+    return t('objectListCountError');
+  }
+
+  return t.rich('objectListCount', {
+    count: objectLists.length,
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/communities/community-card.tsx:54]: Does Clerk have a way of counting the records? This implementation works fine, but for this functionality it looks a bit 'heavy': we retrieve all the records but only need the number of records?

apps/researcher/src/components/add-object-list-form/actions.ts
```diff
@@ -0,0 +1,19 @@
...
+  createdBy: string;
+  name: string;
+  description?: string;
+}
+export async function addList(listItem: ListItem) {
+  const response = await objectList.create(listItem);
+  const community = await getCommunityById(listItem.communityId);
+  revalidatePath(`/[locale]/communities/${community.slug}`, 'page');
+
+  return JSON.stringify(response);
```
sdevalk, line 18: 1. To be sure: do we use this response in other parts of the code?
2. Out of interest: is the `JSON.stringify()` required (i.e. Next doesn't do this for us)?

apps/researcher/src/components/add-object-list-form/actions.ts
```diff
@@ -0,0 +1,19 @@
...
+import {objectList} from '@colonial-collections/database';
+import {revalidatePath} from 'next/cache';
+
+interface ListItem {
+  communityId: string;
+  createdBy: string;
+  name: string;
+  description?: string;
+}
+export async function addList(listItem: ListItem) {
```
sdevalk, line 13: About the names: the function suggests that a 'list' is going to be added, but the argument suggests that a 'list item' is going to be added. Harmonize the names?

apps/researcher/src/components/add-object-list-form/actions.ts
```diff
@@ -0,0 +1,19 @@
...
+import {getCommunityById} from '@/lib/community';
+import {objectList} from '@colonial-collections/database';
+import {revalidatePath} from 'next/cache';
+
+interface ListItem {
+  communityId: string;
+  createdBy: string;
+  name: string;
+  description?: string;
+}
```
sdevalk, line 12: Minor: add a newline after the interface, for readability?

apps/researcher/src/components/add-object-list-form/index.tsx
```diff
@@ -0,0 +1,151 @@
...
+      await addList(listItem);
+      addNotification({
+        id: 'add-object-list-success',
+        message: t.rich('listSuccessfullyAdded', {
+          name: () => <em>{listItem.name}</em>,
+        }),
+        type: 'success',
+      });
+      setIsVisible(slideOutId, false);
+    } catch (error) {
```
sdevalk, line 56: For consistency with other parts of the codebase: use `err` instead?

apps/researcher/src/lib/community.test.ts
```diff
@@ -1,6 +1,40 @@
...
+import {OrganizationMembership} from '@clerk/nextjs/dist/types/server';
+
+const basicOrganization = {
+  id: 'organization1',
+  name: 'Organization 1',
+  slug: 'organization-1',
+  imageUrl: 'https://example.com/image.png',
+  createdAt: 1690000000000,
+  updatedAt: 1690000000000,
+  logoUrl: null,
```
sdevalk, line 13: To be sure: this should indeed be `null`, not `undefined`?

apps/researcher/src/lib/community.test.ts
```diff
@@ -1,6 +1,40 @@
...
-import {isAdmin, Membership, sort, SortBy} from './community';
+import {isAdmin, sort, SortBy} from './community';
+import {OrganizationMembership} from '@clerk/nextjs/dist/types/server';
+
+const basicOrganization = {
+  id: 'organization1',
+  name: 'Organization 1',
+  slug: 'organization-1',
+  imageUrl: 'https://example.com/image.png',
+  createdAt: 1690000000000,
```
sdevalk, line 11: Is there a way to use Date instead (makes the API friendlier)?

apps/researcher/src/messages/nl/messages.json
```diff
@@ -150,8 +152,23 @@
...
+  },
+  "AddObjectListForm": {
+    "title": "Lijst toevoegen",
+    "labelName": "Lijstnaam",
+    "labelDescription": "Beschrijving (optioneel)",
+    "labelDescriptionSubTitle": "Geef een beschrijving over deze lijst. Bijvoorbeeld, wat is de geschiedenis van deze lijst. Vertel iets over het soort objecten.",
+    "buttonSubmit": "Lijst aanmaken",
+    "buttonCancel": "Annuleren",
+    "nameTooSmall": "Lijstnaam is vereist",
+    "nameTooBig": "Lijstnaam mag maximaal 255 tekens bevatten",
```
sdevalk, line 170: Thinking out load: 255 is a restriction of a data type of the database, but from a user perspective this could sound weird: why 255 chars - why not 251 or 260? Can we use a 'sensible' max instead, e.g. 250?

packages/database/src/object-list.ts
```diff
@@ -1,56 +1,64 @@
...
+  withObjects?: boolean;
+  limitObjects?: number;
+}
+
+interface ObjectList extends InferSelectModel<typeof objectLists> {
+  objects?: InferSelectModel<typeof objectItems>[];
 }
 
-async function getCommunityListsWithObjects(communityId: string) {
+export async function getByCommunityId(
```
sdevalk, line 16: Should we validate the incoming options?

sdevalk (commented):
Nice work! I've added a number of questions and suggestions.

apps/researcher/src/lib/community.test.ts
barbarah: No, logo is `string | null`

apps/researcher/src/lib/community.test.ts
barbarah: Only if we build a wrapper around the clerk response. This comes directly from Clerk.

apps/researcher/src/app/[locale]/communities/community-card.tsx
Comment (barbarah) [reply apps/researcher/src/app/[locale]/communities/community-card.tsx]: Good point. This `ObjectListCount` is not Clerk data but from our database. But this is still heavy for our database, so I have added an `objectList.countByCommunityId` function with a SQL `count` function. See commit https://github.com/colonial-heritage/colonial-collections/pull/248/commits/4a53e3443461c4377f0cb7c8eecf3bd526a1c39a

I have also added a `Suspense` around this component so the page can load without all these calls being made.

Above this component, there is also the `MembershipCount` component with a call for all members just for a count of the numbers. This is also heavy; I do this call because the `includeMembersCount` of `clerkClient.organizations.getOrganizationList` does not work. I have asked about this in Discord, and they responded that the fix is in the next release. When this is added, it will be one call per community less.

The caching will also help keep this page not too heavy.

apps/researcher/src/components/add-object-list-form/actions.ts
barbarah: I will first answer 2: This is required, without the `JSON.stringify()`, I get the warning 'Only plain objects can be passed to Client Components from Server Components'. 

However, we do not use the response. So I can also leave this line out.

apps/researcher/src/components/add-object-list-form/actions.ts
barbarah: Good point. This function creates a list. I will rename `listItem` to `list`

packages/database/src/object-list.ts
barbarah: Yes. I even already have a schema for the validation. Validation added in commit: https://github.com/colonial-heritage/colonial-collections/pull/248/commits/034c0980a4d861a480983ff5da551fa53ff9c451

apps/researcher/src/components/add-object-list-form/index.tsx
```diff
@@ -0,0 +1,151 @@
...
+    },
+  });
+
+  const t = useTranslations('AddObjectListForm');
+  const {setIsVisible} = useSlideOut();
+  const {addNotification} = useNotifications();
+
+  const onSubmit: SubmitHandler<FormValues> = async listItem => {
+    try {
+      await addList(listItem);
```
sdevalk, line 47: Or just `list` (to be in sync with apps/researcher/src/components/add-object-list-form/actions.ts)?

packages/database/src/object-list.ts
```diff
@@ -1,56 +1,79 @@
...
   });
 }
 
-interface CreateListForCommunityProps {
+export async function countByCommunityId(communityId: string) {
+  const result = await db
+    .select({count: sql<number>`count(*)`})
+    .from(objectLists)
+    .where(eq(objectLists.communityId, communityId));
+  return result[0].count;
```
sdevalk, line 44: To be sure: is there a scenario where index 0 of `result` does not exist (i.e. should we check for its existence)?

barbarah: I don't think there is a scenario. The Drizzle orm aggregation functions always return an array with one value.

apps/researcher/src/components/add-object-list-form/index.tsx
```diff
@@ -0,0 +1,151 @@
+'use client';
+
+import {useAuth} from '@clerk/nextjs';
+import {addList} from './actions';
+import {useForm, SubmitHandler} from 'react-hook-form';
+import {zodResolver} from '@hookform/resolvers/zod';
+import {insertObjectListSchema} from '@colonial-collections/database/src/db/schema';
```
sdevalk, line 7: Is there a more 'elegant'/abstract way to import from the `database` package? It now contains the full path to a specific place in the `src` folder. Would something like this be possible?

`import {insertObjectListSchema} from '@colonial-collections/database';`

packages/database/src/object-list.ts
sdevalk: Perhaps add a comment explaining this (i.e. we assume that a third-party library always returns an array with one value that is an object with the `count` prop)?

sdevalk (approved):
Nice! Just two small notes.

PR #259, opened 2023-09-29
feat: locations facet

sdevalk:
This PR introduces a new facet for the heritage objects: 'Location'. The facet contains the names of countries in which the objects were made. The name 'Location' is intentionally generic (i.e. it's not called 'Countries'): this allows us to add more information to the facet if needed (e.g. we could also add names of places or continents).

Note: the 'location' (or 'country') is not a part of the details of an object - according to the [design of the search results](https://gui-prototype.colonialcollections.nl/search.html#!) this is not necessary. This information can be added later on, though.


barbarah (commented):
Look good! The end-to-end tests fail because the preview temporarily uses the production environments without the owner filter. I will change this back to the development variables.

PR #266, opened 2023-10-03
feat: facets for 'date created'

sdevalk:
This PR adds two facets to the search functionality of the Research App: `dateCreatedStart` and `dateCreatedEnd`. These are based on the [UI design](https://gui-prototype.colonialcollections.nl/search.html) ('From year', 'Till year').

A note: the facet values consist of years, but the facets have the name 'date'. This is intentional: if we don't want to use years anymore and use some other time unit (e.g. centuries, groups of years, exact dates) then we don't have to rename the facets.

### DIFF
--- a/apps/researcher/src/lib/api/definitions.ts
+++ b/apps/researcher/src/lib/api/definitions.ts
@@ -73,7 +73,11 @@ export enum SortOrder {
 
 export const SortOrderEnum = z.nativeEnum(SortOrder);
 
-export type SearchResultFilter = Thing & {totalCount: number};
+export type SearchResultFilter = {
+  id: string | number;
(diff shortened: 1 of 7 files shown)


barbarah (commented):
If I try to use this filter I get a Zod error.

```
  {
      code: 'invalid_type',
      expected: 'string',
      received: 'array',
      path: [ 'filters', 'dateCreatedStart' ],
      message: 'Expected string, received array'
    }
  ],
```

You can test this by adding '?dateCreatedStart=1725' to the URL.

barbarah (commented):
If I understand correctly, this is only for exact matches. But if I look at the design, we need a range filter. So it returns all objects between `dateCreatedStart` and `dateCreatedEnd`.

barbarah (approved):

PR #267, opened 2023-10-04
Show more filter facets

barbarah:
This pull request shows the newly added facets, except 'date created', because this one needs a custom component. I will add this separately.

Showing these facets (on the real data) raises many questions, and I will ask these questions during the sprint planning.

- There are a lot of materials, too much for the 'show more' functionality. This needs a redesign.
- The filters are now ordered by name. Some options are specific to 1 object. For at least materials ordered by `totalCount` makes more sense.
- There are filters for "Creators", but all options show zero objects. Is this a bug?
- We should do something with the `totalCount`. Showing a number and maybe hiding options with 0 results.


barbarah:
I have talked about this ticket offline. Because most filters have some work needed before they are production-ready, I will close this for now and add a pull request for each filter separately.

barbarah:
I have talked about this ticket offline. Because most filters have some work needed before they are production-ready, I will close this for now and add a pull request for each filter separately.

PR #272, opened 2023-10-06
Add object to list

barbarah:
This pull request will add a dropdown to the object detail page. The user can choose a list to add the object to. After adding the object, the object is shown on the community page.

The design of the object list dropdown I made myself, and maybe needs some changes later.

I made two schema changes:

1: `objectItem.listId` could be null. An `objectItem` without a list makes no sense, so I made this property not null.
2: The `objectItem.objectId` (hashed URI) was the primary key. But objectId is not unique because objects can be added to multiple lists. So, I have added  `objectItem.id` with an auto-increment and made this the primary key.

### DIFF
--- a/apps/researcher/src/app/[locale]/objects/[id]/object-lists-actions.ts
+++ b/apps/researcher/src/app/[locale]/objects/[id]/object-lists-actions.ts
@@ -6,20 +6,25 @@ import {InsertObjectItem} from '@colonial-collections/database';
 import {revalidatePath} from 'next/cache';
 
 export async function getCommunityLists(communityId: string, objectId: string) {
-  return objectList.getByCommunityId(communityId, {includeObjectIri: objectId});
+  return objectList.getByCommunityId(communityId, {objectIri: objectId});
 }
 
 interface AddObjectToListProps {
-  listItem: InsertObjectItem;
+  objectItem: InsertObjectItem;
   communityId: string;
 }
 
 export async function addObjectToList({
-  listItem,
+  objectItem,
   communityId,
 }: AddObjectToListProps) {
-  await objectList.addObject(listItem);
-  const community = await getCommunityBySlug(communityId);
+  const addObjectPromise = objectList.addObject(objectItem);
+  const getCommunityPromise = getCommunityBySlug(communityId);
+  const [community] = await Promise.all([
+    getCommunityPromise,
+    addObjectPromise,
+  ]);
+
   revalidatePath(`/[locale]/communities/${community.slug}`, 'page');
 }
 
--- a/apps/researcher/src/app/[locale]/objects/[id]/object-lists-actions.ts
+++ b/apps/researcher/src/app/[locale]/objects/[id]/object-lists-actions.ts
@@ -5,8 +5,8 @@ import {objectList} from '@colonial-collections/database';
 import {InsertObjectItem} from '@colonial-collections/database';
 import {revalidatePath} from 'next/cache';
 
-export async function getCommunityLists(communityId: string) {
-  return objectList.getByCommunityId(communityId);
+export async function getCommunityLists(communityId: string, objectId: string) {
+  return objectList.getByCommunityId(communityId, {includeObjectIri: objectId});
 }
 
 interface AddObjectToListProps {
@@ -19,6 +19,12 @@ export async function addObjectToList({
   communityId,
 }: AddObjectToListProps) {
   await objectList.addObject(listItem);
+  const community = await getCommunityBySlug(communityId);
+  revalidatePath(`/[locale]/communities/${community.slug}`, 'page');
+}
+
+export async function removeObjectFromList(id: number, communityId: string) {
+  await objectList.removeObject(id);
   const community = await getCommunityBySlug(communityId);
   revalidatePath(`/[locale]/communities/${community.slug}`, 'page');
 }
--- a/apps/researcher/src/app/[locale]/objects/[id]/object-lists-actions.ts
+++ b/apps/researcher/src/app/[locale]/objects/[id]/object-lists-actions.ts
@@ -2,7 +2,7 @@
 
 import {getCommunityBySlug} from '@/lib/community';
 import {objectList} from '@colonial-collections/database';
-import {InsertObjectItem as ObjectItem} from '@colonial-collections/database';
+import {ObjectItemBeingCreated} from '@colonial-collections/database';
 import {revalidatePath} from 'next/cache';
 
 export async function getCommunityLists(communityId: string, objectId: string) {
@@ -10,7 +10,7 @@ export async function getCommunityLists(communityId: string, objectId: string) {
 }
 
 interface AddObjectToListProps {
-  objectItem: ObjectItem;
+  objectItem: ObjectItemBeingCreated;
   communityId: string;
 }
 
--- a/apps/researcher/src/app/[locale]/objects/[id]/object-lists-actions.ts
+++ b/apps/researcher/src/app/[locale]/objects/[id]/object-lists-actions.ts
@@ -2,7 +2,7 @@
 
 import {getCommunityBySlug} from '@/lib/community';
 import {objectList} from '@colonial-collections/database';
-import {InsertObjectItem} from '@colonial-collections/database';
+import {InsertObjectItem as ObjectItem} from '@colonial-collections/database';
 import {revalidatePath} from 'next/cache';
 
 export async function getCommunityLists(communityId: string, objectId: string) {
@@ -10,7 +10,7 @@ export async function getCommunityLists(communityId: string, objectId: string) {
 }
 
 interface AddObjectToListProps {
-  objectItem: InsertObjectItem;
+  objectItem: ObjectItem;
   communityId: string;
 }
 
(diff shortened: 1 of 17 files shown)


barbarah:
@sdevalk, I just removed you as a reviewer. There is a bug in my last changes.

apps/researcher/src/app/[locale]/objects/[id]/object-lists-menu.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+      addNotification({
+        id: 'objectAddedToList',
+        message: t.rich('objectAddedToList', {
+          name: () => <em>{objectList.name}</em>,
+        }),
+        type: 'success',
+      });
+    } catch (err) {
+      if ((err as Error).message.includes('Duplicate entry')) {
```
Comment (barbarah) [inline apps/researcher/src/app/[locale]/objects/[id]/object-lists-menu.tsx:46]: This does not work in production, I will fix this on Monday. The response in production only: 

Error: An error occurred in the Server Components render. The specific message is omitted in production builds to avoid leaking sensitive details. A digest property is included on this error instance which may provide additional details about the nature of the error.

apps/researcher/src/app/[locale]/objects/[id]/object-lists-menu.tsx
```diff
@@ -0,0 +1,168 @@
...
+  communityId,
+  objectId,
+  userId,
+}: CommunityMenuItemsProps) {
+  const [objectLists, setObjectLists] = useState<SelectObjectList[]>([]);
+  const t = useTranslations('ObjectDetails');
+  const {addNotification} = useNotifications();
+
+  async function listClick(objectList: SelectObjectList) {
+    const objectInList = !!objectList.objects!.length;
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/objects/[id]/object-lists-menu.tsx:33]: The name `objectInList`, to me, suggests that we check whether a specific object is in a list. But this line merely seems to check whether there are objects in a list. Rename to e.g. `isEmptyList` to make the meaning clearer?

apps/researcher/src/app/[locale]/objects/[id]/object-lists-menu.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+export default function ObjectListsMenu({objectId}: ObjectListsMenuProps) {
+  const t = useTranslations('ObjectDetails');
+  const {user} = useUser();
+
+  if (!user || !user.organizationMemberships.length) {
+    return null;
+  }
+
+  const communities = user.organizationMemberships.map(m => m.organization);
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/objects/[id]/object-lists-menu.tsx:130]: What's the meaning of `m`? Perhaps use a full name to make this clear?

packages/database/index.ts
sdevalk, line 1: Is it possible to put `index.ts` in the `src` directory, to keep all TS files in one place?

packages/database/src/object-list.ts
```diff
@@ -1,23 +1,22 @@
...
+import {SelectObjectList} from './db/types';
 
 interface Option {
   withObjects?: boolean;
   limitObjects?: number;
-}
-
-interface ObjectList extends InferSelectModel<typeof objectLists> {
-  objects?: InferSelectModel<typeof objectItems>[];
+  includeObjectIri?: string;
```
sdevalk, line 12: Minor: the word `include` could be interpreted as being a boolean (see e.g. JS's `Array.includes()`). Do you see a way to omit the word or replace it with something else?

packages/database/index.ts
barbarah: Not easily. The problem is the following:

I need to spit up the actions that can only be done on the server. And actions that the client can also do. If both client/server actions are exported from one file (`src/index.ts`), and you try to import a client action from a client component, you will still get errors. So they need to be separated.

Suppose I have two export files `src/client` and `scr/index` (for server actions). I must use the `exports` field in the `package.json`. (docs: https://webpack.js.org/guides/package-exports/) to export the two files separately.

But this does not do so well with Typescript. To use this, I need to change the Typescript compiler setting `module` to `NodeNext` in the app, which gives many more problems. (If you want more info, I must recheck this).

So, most external packages compile the Typescript to JavaScript in a `/dist` folder and export from the `/dist` folder. In pure JavaScript, the `exports` field in the `package.json` does work. See, for example, [`next-intl`](https://github.com/amannn/next-intl/blob/main/packages/next-intl/package.json#L26C4-L26C11). This package also has a separate export file for server actions, exported from the `/dist` folder.

I have not tried this yet. Moving the export files to the root was the easiest way to have multiple export points without introducing an extra build step. If the export files are in the root I do not need the `exports` field in the `package.json`.

Maybe the `/dist` approach is the best option. I already spent hours figuring out the above. So, I decided on the easiest working option.

apps/researcher/src/app/[locale]/objects/[id]/object-lists-actions.ts
```diff
@@ -0,0 +1,30 @@
...
+import {objectList} from '@colonial-collections/database';
+import {InsertObjectItem} from '@colonial-collections/database';
+import {revalidatePath} from 'next/cache';
+
+export async function getCommunityLists(communityId: string, objectId: string) {
+  return objectList.getByCommunityId(communityId, {includeObjectIri: objectId});
+}
+
+interface AddObjectToListProps {
+  listItem: InsertObjectItem;
```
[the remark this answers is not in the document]
Comment (barbarah) [reply apps/researcher/src/app/[locale]/objects/[id]/object-lists-actions.ts:13]: I have renamed `listItem` to `objectItem`, I think this is more clear.

packages/database/src/object-list.ts
barbarah: I will just call it `objectIri`

packages/database/index.ts
barbarah: I would like to fix this, but not in this pull request.

sdevalk: Thanks for the clarification! Agree, let's not fix this in this PR but in a separate one.

I'm all in favor of using ESM instead of CommonJS, by the way (e.g. `module` = `nodenext`).

apps/researcher/src/app/[locale]/objects/[id]/object-lists-actions.ts
```diff
@@ -0,0 +1,35 @@
...
+import {objectList} from '@colonial-collections/database';
+import {InsertObjectItem} from '@colonial-collections/database';
+import {revalidatePath} from 'next/cache';
+
+export async function getCommunityLists(communityId: string, objectId: string) {
+  return objectList.getByCommunityId(communityId, {objectIri: objectId});
+}
+
+interface AddObjectToListProps {
+  objectItem: InsertObjectItem;
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/objects/[id]/object-lists-actions.ts:13]: Naming-wise this is still a bit confusing to me - I would expect an 'object item' to be of type 'ObjectItem' instead of 'InsertObjectItem'. Do you see a way to change this?

apps/researcher/src/app/[locale]/objects/[id]/object-lists-menu.tsx
```diff
@@ -0,0 +1,170 @@
...
+        enterFrom="transform opacity-0 scale-95"
+        enterTo="transform opacity-100 scale-100"
+        leave="transition ease-in duration-75"
+        leaveFrom="transform opacity-100 scale-100"
+        leaveTo="transform opacity-0 scale-95"
+      >
+        <Menu.Items className="absolute right-0 z-10 mt-2 w-56 origin-top-right divide-y divide-gray-100 rounded-md bg-white shadow-lg ring-1 ring-black ring-opacity-5 focus:outline-none">
+          {communities.map(community => (
+            <div key={community.id}>
+              <div className="font-semibold	 px-2 pt-2">{community.name}</div>
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/objects/[id]/object-lists-menu.tsx:158]: Minor: there is some excessive whitespace here :)

apps/researcher/src/app/[locale]/objects/[id]/object-lists-actions.ts
Comment (barbarah) [reply apps/researcher/src/app/[locale]/objects/[id]/object-lists-actions.ts:13]: I can rename it on line 5. However, I would like to keep the types `InsertObjectItem` and `SelectObjectItem`. Because for the insert `id` and `createdAt` are optional. For the `SelectObjectItem` it is not.

Comment (barbarah) [reply apps/researcher/src/app/[locale]/objects/[id]/object-lists-actions.ts]: See commit: https://github.com/colonial-heritage/colonial-collections/pull/272/commits/7562b4565938830ddef71d51fa277023f195b193

Is this ok?

Comment (sdevalk) [reply apps/researcher/src/app/[locale]/objects/[id]/object-lists-actions.ts]: An alias, that's a creative solution :) But I don't think it solves the real issue, though: the naming and typing of things.

Currently the types originate from the `database` package and are used in the frontend. This leads to tight-coupling. For example: the names `InsertObjectItem` and `SelectObjectItem` are 'database terminology'. But I think the frontend should be able to use types that are storage-agnostic - hence the generic entity name `ObjectItem`. (The fact that the database package uses it to insert or to select an object item is an implementation detail then.)

Yes, I fully agree that you want to keep different types for insertion and selection. But from an API perspective I would use more generic names that identify the nature/state of the object item. For example: `ObjectItem` for a full object item with all properties and `ObjectItemBeingCreated` for an object item that doesn't have all props yet.

Having said all this, feel free to not implement this, or not now (the ~~show~~ PR must go on).

sdevalk (approved):

PR #274, opened 2023-10-06
Community description

barbarah:
In this pull request, the communities get descriptions. You can see the description on the community overview and details page. The community is saved as pubic metadata in Clerk.

Things to note:

- I have also added a custom type for Community and Memberships in this pull request. As discussed in previous pull requests.
- I had to set the community page to `export const dynamic = 'force-dynamic';` ([docs](https://nextjs.org/docs/app/api-reference/file-conventions/route-segment-config#dynamic)). The Clerk requests were cached, and updated descriptions or member lists were not visible.
- I have set the max length of the description to 2000. Metadata is limited to 8kb maximum, so there is still room for other properties if we need it.

### DIFF
--- a/apps/researcher/src/app/[locale]/communities/[slug]/actions.ts
+++ b/apps/researcher/src/app/[locale]/communities/[slug]/actions.ts
@@ -3,21 +3,34 @@
 import {joinCommunity, updateDescription} from '@/lib/community';
 import {revalidatePath} from 'next/cache';
 
-interface UpdateDescriptionActionProps {
+interface UpdateDescriptionAndRevalidateProps {
   communityId: string;
   communitySlug: string;
   description: string;
 }
 
-export async function updateDescriptionAction({
+export async function updateDescriptionAndRevalidate({
   communityId,
   communitySlug,
   description,
-}: UpdateDescriptionActionProps) {
+}: UpdateDescriptionAndRevalidateProps) {
   await updateDescription({communityId, description});
   revalidatePath(`/[locale]/communities/${communitySlug}`, 'page');
   revalidatePath('/[locale]/communities', 'page');
 }
 
-// Export as server actions.
-export {joinCommunity as joinCommunityAction};
+interface JoinCommunityAndRevalidateProps {
+  communityId: string;
+  communitySlug: string;
+  userId: string;
+}
+
+export async function joinCommunityAndRevalidate({
+  communityId,
+  communitySlug,
+  userId,
+}: JoinCommunityAndRevalidateProps) {
+  await joinCommunity({communityId, userId});
+  revalidatePath(`/[locale]/communities/${communitySlug}`, 'page');
+  revalidatePath('/[locale]/communities', 'page');
+}
--- a/apps/researcher/src/app/[locale]/communities/[slug]/actions.ts
+++ b/apps/researcher/src/app/[locale]/communities/[slug]/actions.ts
@@ -1,6 +1,23 @@
 'use server';
 
-import {joinCommunity} from '@/lib/community';
+import {joinCommunity, editDescription} from '@/lib/community';
+import {revalidatePath} from 'next/cache';
+
+interface editDescriptionActionProps {
+  communityId: string;
+  communitySlug: string;
+  description: string;
+}
+
+async function editDescriptionAction({
+  communityId,
+  communitySlug,
+  description,
+}: editDescriptionActionProps) {
+  await editDescription({communityId, description});
+  revalidatePath(`/[locale]/communities/${communitySlug}`, 'page');
+  revalidatePath('/[locale]/communities', 'page');
+}
 
 // Export as server actions.
-export {joinCommunity};
+export {joinCommunity, editDescriptionAction};
--- a/apps/researcher/src/app/[locale]/communities/[slug]/actions.ts
+++ b/apps/researcher/src/app/[locale]/communities/[slug]/actions.ts
@@ -20,4 +20,4 @@ export async function updateDescriptionAction({
 }
 
 // Export as server actions.
-export {joinCommunity};
+export {joinCommunity as joinCommunityAction};
--- a/apps/researcher/src/app/[locale]/communities/[slug]/actions.ts
+++ b/apps/researcher/src/app/[locale]/communities/[slug]/actions.ts
@@ -1,23 +1,23 @@
 'use server';
 
-import {joinCommunity, editDescription} from '@/lib/community';
+import {joinCommunity, updateDescription} from '@/lib/community';
 import {revalidatePath} from 'next/cache';
 
-interface editDescriptionActionProps {
+interface UpdateDescriptionActionProps {
   communityId: string;
   communitySlug: string;
   description: string;
 }
 
-async function editDescriptionAction({
+export async function updateDescriptionAction({
   communityId,
   communitySlug,
   description,
-}: editDescriptionActionProps) {
-  await editDescription({communityId, description});
+}: UpdateDescriptionActionProps) {
+  await updateDescription({communityId, description});
   revalidatePath(`/[locale]/communities/${communitySlug}`, 'page');
   revalidatePath('/[locale]/communities', 'page');
 }
 
 // Export as server actions.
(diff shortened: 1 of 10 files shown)


apps/researcher/src/lib/community.ts
```diff
@@ -1,24 +1,81 @@
...
+    createdAt: organization.createdAt,
+  };
+}
+
+function organizationMembershipToCommunityMembership(
+  membership: OrganizationMembership
+): Membership {
+  if (!membership.publicUserData) {
+    throw new Error(
+      'organizationMembershipToCommunityMembership: publicUserData is undefined'
```
sdevalk, line 39: Should we add the name of the function to the error message? If this error is thrown, the error stack should make clear where the error comes from (e.g. from which function), so adding the name feels redundant?

apps/researcher/src/lib/community.ts
```diff
@@ -104,3 +158,17 @@ export async function joinCommunity({
     role: 'basic_member',
   });
 }
+
+export async function editDescription({
```
sdevalk, line 162: The term 'edit', to me, doesn't necessarily mean that data is going to be stored. Change the term to e.g. 'save' or 'update'?

apps/researcher/src/messages/en/messages.json
```diff
@@ -158,7 +158,14 @@
     "objectListsSubTitle": "This community has {count, plural, =0 {0 lists} =1 {1 list} other {# lists}}",
     "membersTitle": "Members",
     "error": "Something went wrong while fetching the community. Please try again later.",
-    "addObjectListButton": "Add list"
+    "addObjectListButton": "Add list",
+    "editDescriptionButton": "Edit description",
+    "labelDescription":"Description",
+    "changeDescriptionSaveButton": "Save",
+    "changeDescriptionCancelButton": "Cancel",
+    "descriptionSuccessfullyEdited": "Description has been updated",
```
sdevalk, line 166: Or: `descriptionSaved` or `descriptionUpdated`? The edit could have been successful, but it's actually about storing the description that has been edited?

apps/researcher/src/messages/nl/messages.json
```diff
@@ -158,7 +158,14 @@
     "objectListsSubTitle": "Deze community heeft {count, plural, =0 {0 lijsten} =1 {1 lijst} other {# lijsten}}",
     "membersTitle": "Leden",
     "error": "Er is een fout opgetreden bij het ophalen van de community. Probeer het later opnieuw.",
-    "addObjectListButton": "Lijst toevoegen"
+    "addObjectListButton": "Lijst toevoegen",
+    "editDescriptionButton": "Beschrijving aanpassen",
+    "labelDescription":"Beschrijving",
+    "changeDescriptionSaveButton": "Opslaan",
+    "changeDescriptionCancelButton": "Annuleren",
+    "descriptionSuccessfullyEdited": "Beschrijving is succesvol aangepast",
```
sdevalk, line 166: Perhaps omit 'succesvol' - success is an implication of 'aangepast'?

apps/researcher/src/app/[locale]/communities/[slug]/actions.ts
```diff
@@ -1,6 +1,23 @@
...
+import {joinCommunity, updateDescription} from '@/lib/community';
+import {revalidatePath} from 'next/cache';
+
+interface UpdateDescriptionActionProps {
+  communityId: string;
+  communitySlug: string;
+  description: string;
+}
+
+export async function updateDescriptionAction({
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/communities/[slug]/actions.ts:12]: If I understand this correctly, there are two server actions in this file, on line 12 and on line 23. Their names differ, though: the first one uses the `Action` suffix. Is there a way to harmonize these names?

apps/researcher/src/lib/community.ts
```diff
@@ -1,24 +1,79 @@
...
+    imageUrl: organization.imageUrl,
+    createdAt: organization.createdAt,
+  };
+}
+
+function organizationMembershipToCommunityMembership(
+  membership: OrganizationMembership
+): Membership {
+  if (!membership.publicUserData) {
+    throw new Error('publicUserData is undefined');
```
sdevalk, line 38: Minor: this throws when `publicUserData` is undefined. But if it is defined, lines 44 to 48 assume that it is of a certain type, e.g. it has props such as `userId`. Is this always the case? If so, make this clear in a comment? If this is not the case, validate the `publicUserData` with zod?

apps/researcher/src/app/[locale]/communities/[slug]/actions.ts
Comment (barbarah) [reply apps/researcher/src/app/[locale]/communities/[slug]/actions.ts:12]: Yes, this is correct. I will use the suffix `Action` for both actions and differentiate it from the functions in '@/lib/community'.

apps/researcher/src/lib/community.ts
barbarah: I have rewritten this function a bit. I have set all my assumptions about the Clerk data in a code comment and let typescript throw errors by adding '!' to the code.

apps/researcher/src/app/[locale]/communities/[slug]/actions.ts
Comment (barbarah) [reply apps/researcher/src/app/[locale]/communities/[slug]/actions.ts:12]: See: https://github.com/colonial-heritage/colonial-collections/pull/274/commits/e4243df16d1b52fc81e64d966616d2a70d8dd81a

Comment (sdevalk) [reply apps/researcher/src/app/[locale]/communities/[slug]/actions.ts:12]: Thanks! Some considerations:

1. The `Action` suffix is somewhat redundant because the functions are in a file called `actions.ts` that make the scope of the functions clear?
2. You could argue that all functions are 'actions', so adding the suffix `Action` doesn't say that much?
3. Consistency: if we add the suffix `Actions` to server actions, should we also add the suffix `Component` to components?

Does Next.js have a convention for naming server actions that we can follow?

Comment (barbarah) [reply apps/researcher/src/app/[locale]/communities/[slug]/actions.ts:12]: Next.js does not have a convention for naming server actions. In their docs, I also see some actions with the suffix 'Action' and some without.

I added the suffix so it is clear that it was different from the functions in '@/lib/community' with similar names. But then again the file name (`actions.ts`) also makes that clear. I don't really have an opinion about this subject. I will remove the suffix 'Action'.

Comment (barbarah) [reply apps/researcher/src/app/[locale]/communities/[slug]/actions.ts:12]: I looked at this file some more. `joinCommunity` was one of the first server actions. But this function should also revalidate pages, without this you won't see your avatar in the member list, and the member count can be wrong. Also, it has `organizationId` as input in the rest of the community logic I use the term `communityId`. So I made some changes to this action.

Comment (barbarah) [reply apps/researcher/src/app/[locale]/communities/[slug]/actions.ts]: The code is updated. I named the actions `updateDescriptionAndRevalidate` and `joinCommunityAndRevalidate`

apps/researcher/src/app/[locale]/communities/[slug]/page.tsx
```diff
@@ -10,7 +10,8 @@ import {revalidatePath} from 'next/cache';
 import {objectList} from '@colonial-collections/database';
 import ObjectCard from './object';
 import AddObjectListForm from '@/components/add-object-list-form';
-import {SlideOutButton, SlideOut, Notifications} from 'ui';
+import {SlideOutButton, SlideOut, SlideOutClosed, Notifications} from 'ui';
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/communities/[slug]/page.tsx:13]: Out of curiosity: we make use of the `database` package by importing from `@colonial-collections/database` but we make use of the `ui` package by importing from `ui` (without the `@colonial-collections` prefix). Is there a rule that we should follow?

sdevalk (approved):

apps/researcher/src/app/[locale]/communities/[slug]/page.tsx
Comment (barbarah) [reply apps/researcher/src/app/[locale]/communities/[slug]/page.tsx:13]: This is because this was the first package I made, and we did not have this convention then. But I can update it so they are all the same.

Comment (sdevalk) [reply apps/researcher/src/app/[locale]/communities/[slug]/page.tsx:13]: That sounds perfect! Doesn't have to be a part of this PR, though.

PR #280, opened 2023-10-12
Update logic search-params based on updated assumptions

barbarah:
Now that more filters are added to the API, I noticed I needed to correct some assumptions. I thought:

- Filter IDs never contain commas
- Filter IDs are always strings (the dates have number ids)
- Filters are always arrays (the dates have a single value)
- There is only one type of facet component (for example, no date component). This was not really a wrong assumption, but the code was not ready for different types of components.

This pull request tries to fix all the wrong assumptions.

Instead of separating the search params with commas. Different values are separated with the format '?key=a&key=b'.

Note: some applications use the format '?key[]=a&key[]=b'. But this depends on the programming language. PHP and C# will automatically parse the format '?key[]=a&key[]=b'. But Next.js uses the javascript function [`URLSearchParams: getAll()`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/getAll) under the hood. So it will automatically parse the format '?key=a&key=b'' to `['a', 'b']`.

There are two things to be aware of:

1. The parsed search params from Next.Js will show a string if there is only one value and an array if there are multiple values. For example, '?foo=a&foo=2&bar=3' will parse to: `{foo: ['1', '2'], bar: '3'}`.
2. All values are strings.

Depending on the facets, we want to transform the Next.js parsed search params to arrays, strings or numbers.

### DIFF
--- a/apps/researcher/src/app/[locale]/page.tsx
+++ b/apps/researcher/src/app/[locale]/page.tsx
@@ -104,7 +104,6 @@ export default async function Home({searchParams = {}}: Props) {
   try {
     searchResult = await heritageObjects.search(searchOptions);
   } catch (err) {
-    console.log(err);
     hasError = true;
     console.error(err);
   }
--- a/apps/researcher/src/app/[locale]/page.tsx
+++ b/apps/researcher/src/app/[locale]/page.tsx
@@ -8,6 +8,7 @@ import {
   fromSearchParamsToSearchOptions,
   getClientSortBy,
   defaultSortBy,
+  SearchParamTypes,
 } from '@colonial-collections/list-store';
 import {
   SearchResult,
@@ -26,38 +27,43 @@ import {
 import {SmallScreenSubMenu, SubMenuButton, SubMenuDialog} from 'ui';
 import {AdjustmentsHorizontalIcon} from '@heroicons/react/20/solid';
 import Tabs from './tabs';
+import {ElementType} from 'react';
 
 // Revalidate the page every n seconds
 export const revalidate = 60;
 
-// Set the order of the filters
-const filterKeysOrder: ReadonlyArray<keyof SearchResult['filters']> = [
-  'owners',
-  'types',
-  'subjects',
-  'publishers',
+interface FacetProps {
+  name: keyof SearchResult['filters'];
+  searchParamType: SearchParamTypes;
+  Component: ElementType;
+}
+
+const facets: ReadonlyArray<FacetProps> = [
+  {name: 'owners', searchParamType: 'array', Component: FilterSet},
+  {name: 'types', searchParamType: 'array', Component: FilterSet},
+  {name: 'subjects', searchParamType: 'array', Component: FilterSet},
+  {name: 'publishers', searchParamType: 'array', Component: FilterSet},
 ];
 
 interface FacetMenuProps {
   filters: SearchResult['filters'];
-  filterKeysOrder: ReadonlyArray<keyof SearchResult['filters']>;
 }
 
-function FacetMenu({filterKeysOrder, filters}: FacetMenuProps) {
+function FacetMenu({filters}: FacetMenuProps) {
   const t = useTranslations('Filters');
 
   return (
     <>
       <SearchFieldWithLabel />
-      {filterKeysOrder.map(
-        filterKey =>
-          !!filters[filterKey]?.length && (
-            <FilterSet
-              key={filterKey}
-              title={t(`${filterKey}Filter`)}
-              searchResultFilters={filters[filterKey]}
-              filterKey={filterKey}
-              testId={`${filterKey}Filter`}
+      {facets.map(
+        ({name, Component}) =>
+          !!filters[name]?.length && (
+            <Component
+              key={name}
+              title={t(`${name}Filter`)}
+              searchResultFilters={filters[name]}
+              filterKey={name}
+              testId={`${name}Filter`}
             />
           )
       )}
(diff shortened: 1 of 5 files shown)


packages/list-store/src/search-params.test.ts
```diff
@@ -36,6 +37,18 @@ const sortMapping = {
   },
 };
 
+interface FilterKey {
+  name: string;
+  searchParamType: SearchParamTypes;
+}
```
sdevalk, line 43: I'm not used to creating interfaces in tests (I normally test interfaces in tests). Is there a way to create the interface in the actual code and use it in the test?

packages/list-store/src/search-params.ts
```diff
@@ -73,16 +94,11 @@ function fallback<T>(value: T) {
...
-  .string()
-  .optional()
-  .transform(filterValue => filterValue?.split(',').filter(id => !!id))
-  .pipe(z.array(z.string()).optional().default([]));
-
 interface FromSearchParamsToSearchOptionsProps {
   // `searchParams` is a key value object based on the url search params.
   searchParams: {
-    [filter: string]: string;
+    // Search params are always strings or array of strings. Note that numbers are also strings.
```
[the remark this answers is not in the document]
barbarah: I have updated the comment and removed the type 'undefined'.

packages/list-store/src/search-params.ts
```diff
@@ -101,6 +117,10 @@ interface FromSearchParamsToSearchOptionsProps {
       };
     };
   };
+  filterKeys: {
+    name: string;
+    searchParamType: SearchParamTypes;
```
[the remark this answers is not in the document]
barbarah: I have renamed it to `type`

packages/list-store/src/search-params.ts
```diff
@@ -55,11 +59,30 @@ export function getUrlWithSearchParams({
     });
   }
 
-  // Only add relevant values to the search params. Remove all keys with empty strings as values
-  Object.keys(searchParams).forEach(key =>
-    searchParams[key] === '' ? delete searchParams[key] : {}
-  );
-  const encodedSearchParams = new URLSearchParams(searchParams).toString();
+  // Only add relevant values to the search params.
```
sdevalk, line 62: Very minor: the comment suggests that relevant values are going to be added. But lines 64-66 do not add values, only remove. Rephrase the comment?

packages/list-store/src/search-params.ts
```diff
@@ -55,11 +59,30 @@ export function getUrlWithSearchParams({
...
+  Object.keys(searchParams).forEach(key => {
+    if (searchParams[key] === '' || searchParams[key].length === 0) {
```
sdevalk, line 64: Not sure whether this is an option, but is there a way to make these checks a part of the transformations on lines 72-83, so as to make the code more compact/sit together?

packages/list-store/src/search-params.ts
```diff
@@ -73,16 +96,11 @@ function fallback<T>(value: T) {
...
-  .optional()
-  .transform(filterValue => filterValue?.split(',').filter(id => !!id))
-  .pipe(z.array(z.string()).optional().default([]));
-
-interface FromSearchParamsToSearchOptionsProps {
+export interface FromSearchParamsToSearchOptionsProps {
   // `searchParams` is a key value object based on the url search params.
   searchParams: {
-    [filter: string]: string;
+    // Search params are always strings or arrays of strings. Note that numbers are also strings, so we need to transform them to numbers.
```
sdevalk, line 102: I find the addition `so we need to transform them to numbers.` a little bit confusing - the code doesn't transform strings to numbers? Would it be an option to remove the comment as a whole (because the type definition on line 103 already shows the types allowed)?

barbarah: The transformation is done on line 166. I thought it would be relevant to mention that even if the input is always strings, the output is not. But I will remove this line.


The transformation code. `+value` transforms a string to a number.
```
          } else if (filterKey.type === 'number') {
            acc[filterKey.name] = z
              .string()
              .transform(value => +value)
              .optional()
              .or(fallback(undefined));
          }
```

packages/list-store/src/search-params.ts
barbarah: Yes, with the new code, this makes more sense.

apps/researcher/src/app/[locale]/page.tsx
```diff
@@ -93,6 +104,7 @@ export default async function Home({searchParams = {}}: Props) {
   try {
     searchResult = await heritageObjects.search(searchOptions);
   } catch (err) {
+    console.log(err);
     hasError = true;
     console.error(err);
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/page.tsx:109]: Remove this line?

packages/list-store/src/search-params.ts
```diff
@@ -73,16 +89,10 @@ function fallback<T>(value: T) {
...
 
-const searchOptionsFilterSchema = z
-  .string()
-  .optional()
-  .transform(filterValue => filterValue?.split(',').filter(id => !!id))
-  .pipe(z.array(z.string()).optional().default([]));
-
-interface FromSearchParamsToSearchOptionsProps {
+export interface FromSearchParamsToSearchOptionsProps {
   // `searchParams` is a key value object based on the url search params.
```
sdevalk, line 93: Very, very minor: `key-value object` (hyphen) and `URLs` (uppercase)?

sdevalk (approved):
Nice! I've added two small comments, but let's merge this PR otherwise :)

PR #282, opened 2023-10-16
fix: required object identifier

sdevalk:
This PR fixes an issue that appeared in the new data that we received from NMVW last Friday: apparently there can be objects without an identifier, whereas the search engine had always assumed there would be one.

This PR, in the slipstream, also fixes some integration tests that failed due to changes that have been made to the mock data in the testing environment.


barbarah (approved):
Looks good. 

In the front end, the identifier is only shown in the top right of the object details page. If this is empty, it will show: 'id:'. This may need to be updated, but this is not blocking. I will mention it to Bas.

sdevalk:
> Looks good.
> 
> In the front end, the identifier is only shown in the top right of the object details page. If this is empty, it will show: 'id:'. This may need to be updated, but this is not blocking. I will mention it to Bas.

Perfect, thank you!

PR #288, opened 2023-10-18
Searchable multi select facets

barbarah:
This is the new searchable multi-select facet.

There are still discussions about `number_count` and maybe a different facet for materials. So, this is just the first version.


apps/researcher/src/messages/en/messages.json
```diff
@@ -109,14 +109,16 @@
...
-    "ownersFilter": "Owners",
     "typesFilter": "Types",
     "subjectsFilter": "Subjects",
     "birthYearsFilter": "Birth years",
     "birthPlacesFilter": "Birth places",
     "deathYearsFilter": "Death years",
     "deathPlacesFilter": "Death places",
     "publishersFilter": "Object data providers",
+    "materialsFilter": "Materials",
+    "creatorsFilter": "Creators",
```
sdevalk, line 120: Minor: the Object ID standard calls these 'Makers'. Rename the label?

apps/researcher/src/messages/en/messages.json
```diff
@@ -109,14 +109,16 @@
...
+    "locationsFilter": "Locations",
```
sdevalk, line 121: 'Locations' as a label is generic. For example: is this the location of creation, the location where an object currently is, or the location depicted on an object? Is it possible to rename the label and be explicit about its meaning?

packages/list-store/src/use-searchable-facet.tsx
```diff
@@ -0,0 +1,227 @@
...
+
+export function extendFiltersWithLetterCategory(filters?: Filter[]) {
+  if (!filters?.length) {
+    return [];
+  }
+  return filters.map(filter => {
+    let letterCategory = '';
+    const name = filter.name?.toString();
+    if (name) {
+      const firstLetter = name.toUpperCase().replace(/[^0-9A-Z]/g, '')[0];
```
sdevalk, line 117: 1. Very minor: instead of removing _all_ chars that don't meet the criteria, you could just remove the _leading_ chars; makes the regex execution perhaps more efficient.
2. Curious: can `name` be empty? If so, what happens if you pick the first char with `[0]`?

packages/list-store/src/use-searchable-facet.tsx
```diff
@@ -0,0 +1,227 @@
...
+    filtersWithLetterCategory,
+    searchValue
+  );
+
+  const filtersFilteredByLetterCategory = filterByLetterCategory(
+    filtersFilteredBySearchValue,
+    letterCategory
+  );
+
+  if (sortBy === 'alphabetical') {
```
sdevalk, line 175: Is this `alphabetical` the same as the one in `FacetSortBy`? If so, use the latter?

packages/list-store/src/use-searchable-facet.tsx
```diff
@@ -0,0 +1,227 @@
...
+  } else {
+    return filtersFilteredByLetterCategory.sort((a, b) => {
+      return b.totalCount - a.totalCount;
+    });
+  }
+}
+
+export function useSearchableMultiSelectFacet() {
+  const storeContext = useContext(FacetContext);
+  if (!storeContext) throw new Error('Missing FacetContext in the tree');
```
sdevalk, line 191: This is perfectly valid code, but for consistency with other parts of the codebase I prefer the use of curly brackets:

```
if (!storeContext) {
   // ...
}
```

packages/list-store/src/use-searchable-facet.tsx
```diff
@@ -0,0 +1,227 @@
...
+
+function filterBySearchValue(
+  filters: SearchableFilter[],
+  searchValue?: string
+) {
+  if (searchValue) {
+    return filters.filter(
+      filter => filter.name?.toLowerCase().includes(searchValue.toLowerCase())
+    );
+  } else return filters;
```
sdevalk, line 147: This is perfectly valid code, but for consistency with other parts of the codebase I prefer the use of curly brackets:

```
} else {
   return filters;
}
```

packages/list-store/src/use-searchable-facet.tsx
```diff
@@ -0,0 +1,227 @@
...
+  searchValue: string;
+  letterCategory: string;
+  sortBy: FacetSortBy;
+}
+
+function filterBySearchValue(
+  filters: SearchableFilter[],
+  searchValue?: string
+) {
+  if (searchValue) {
```
sdevalk, line 143: To avoid the `if... else...` you could do:

```
if (!searchValue) {
  return filters;
}

return filters.filter(
  filter => filter.name?.toLowerCase().includes(searchValue.toLowerCase())
);
```

packages/list-store/src/use-searchable-facet.tsx
```diff
@@ -0,0 +1,227 @@
...
+
+  const store = useStore(storeContext);
+  const {searchValue, letterCategory, sortBy, filters} = store;
+
+  const filtersWithLetterCategory = useMemo(
+    () => extendFiltersWithLetterCategory(filters),
+    [filters]
+  );
+
+  console.log(filters);
```
sdevalk, line 201: Remove this one?

packages/ui/list/multi-select-facet.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+interface Props {
+  title: string;
+  filters: SearchResultFilter[];
+  filterKey: string;
+  testId?: string;
+}
+
+export function MultiSelectFacet({title, filters, filterKey, testId}: Props) {
+  if (!filters.length) return null;
```
sdevalk, line 14: For consistency, use curly brackets?

sdevalk (commented):
Great work!

apps/researcher/src/messages/en/messages.json
barbarah: This name depends on the filters in the `SearchResult`. The `SearchResult` contains `creators` and not `makers`. Is it possible to change this in the API?

apps/researcher/src/messages/en/messages.json
barbarah: Now, these names are built by adding 'Filter' to the `SearchResult` filter key. In the `SearchResult`, this key is called `location`. Maybe I could make this name customizable instead of building them depending on the `SearchResult` filter key. Or should the API also return a name that is more explicit about its meaning?

apps/researcher/src/messages/en/messages.json
sdevalk: Do you mean that the value `Creators` (not the key `creatorsFilter`) depends on something else?

apps/researcher/src/messages/en/messages.json
sdevalk: 1. Good idea to change the key `locations` in the API - I'll write this down
2. Regardless of (1): I think we should not have a tight coupling between the keys that the API uses and values that we present to users. But perhaps I don't understand your comment? We could ask Bas to suggest a value for `locationsFilter`? If `Locations` is the label he wants, I'm fine with that (it's a display issue) :)

barbarah: Sorry, with name. I meant the translation key. In `apps/researcher/src/app/[locale]/page.tsx` this is the code for the `FacetMenu`:

```
function FacetMenu({filters}: FacetMenuProps) {
  const t = useTranslations('Filters');

  return (
      .....
      {facets.map(({key, Component, ...customProps}) => {
       ....
        return (
          <Component
            key={key}
            title={t(`${key}Filter`)}
            testId={`${key}Filter`}
          />
        );
      })}
    ....
  );
}
```

Where title is ` title={t(`${name}Filter`)}`. This key is the same key used in `SearchResult`.

apps/researcher/src/messages/en/messages.json
barbarah: Sorry, here you were referring to the value and not the translation key. I will change the value. I first thought you wanted the translation key to be `makersFilter`

apps/researcher/src/messages/en/messages.json
barbarah: I'm sorry; I see now that you were also referring to the actual value and not the translation key. This is a lot easier. I will change the value.

packages/list-store/src/use-searchable-facet.tsx
barbarah: Do you mean `if (sortBy === FacetSortBy.alphabetical) {`?

For that, `FacetSortBy` needs to be an `enum`. This was a literal type, not an enum. I don't really like typescript enums for different reasons. But we are probably too far in this project to start this discussion. So, I will change it to an enum.

packages/list-store/src/use-searchable-facet.tsx
barbarah: 1: 👍 

2: The `if (name) {` on line 116 prevents empty names. But `name.toUpperCase().replace(/[^0-9A-Z]/g, '')` could still result in an empty string. `''[0]` will result in `undefined`.

If `letterCategory` is `undefined`, it would be ignored by the function `getLetterCategories`. So filters with the name `undefined`, "", "$%@" cannot be found in the letter category filter, I think this is ok.

I have updated the code a little bit. `letterCategory` is now optional, and I have added tests with the names `undefined`, "" and "#".

packages/list-store/src/use-searchable-facet.tsx
sdevalk: Ok! But do you still want to do a `if (sortBy === FacetSortBy.alphabetical)`?

sdevalk (approved):

packages/list-store/src/use-searchable-facet.tsx
barbarah: Sorry, I thought I had done that!

PR #291, opened 2023-10-20
Update package name 'ui' to '@colonial-collections/ui'

barbarah:
The 'ui' package did not have the prefix `@colonial-collections`. Also, the package should have been included in the apps package.json. Now, it has the same convention as the other packages.

### DIFF
--- a/apps/researcher/src/app/[locale]/objects/[id]/page.tsx
+++ b/apps/researcher/src/app/[locale]/objects/[id]/page.tsx
@@ -14,13 +14,13 @@ import {
   SlideOverHeader,
   SlideOverContent,
   SlideOver,
+  Notifications,
 } from '@colonial-collections/ui';
 import useCurrentPublisher from './useCurrentPublisher';
 import {env} from 'node:process';
 import {formatDateCreated} from './format-date-created';
 import ObjectListsMenu from './object-lists-menu';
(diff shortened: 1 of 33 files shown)


barbarah:
The end-to-end tests fail because the preview environment temporarily uses the production data. The production data does not have an 'owner' filter. I will change these failing tests to use a different filter in a different pull request.

apps/researcher/src/app/[locale]/objects/[id]/page.tsx
```diff
@@ -14,13 +14,13 @@ import {
...
   SlideOver,
-} from 'ui';
+} from '@colonial-collections/ui';
 import useCurrentPublisher from './useCurrentPublisher';
 import {env} from 'node:process';
 import {formatDateCreated} from './format-date-created';
 import ObjectListsMenu from './object-lists-menu';
 import {SignedIn} from '@clerk/nextjs';
-import {Notifications} from 'ui';
+import {Notifications} from '@colonial-collections/ui';
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/objects/[id]/page.tsx:23]: There are several imports from `@colonial-collections/ui` here. Would it be possible to use one import instead, to make visually clear which functions come from this package? If not, place the imports from the same package together?

sdevalk (commented):

sdevalk (approved):
Perfect! :)

PR #292, opened 2023-10-20
Bugfix: Don't scroll to the top of the page after filter selection

barbarah:
Bugfix: Don't scroll to the top of the page after filter selection.

## Current behaviour on the [Production app](https://app.colonialcollections.nl/):

1. Scroll down
3. Select a filter
4. After the filter is added to the URL, the page will scroll to the top of the page

## Wanted behaviour:

The scroll position doesn't change after selecting a filter.

## The fix

There have been some changes in router logic in Next. We must now add the option `{scroll: false}` to the `router.replace()`. We don't use the router directly from next but we use the router of `next-intl`.  `next-intl` needed to be updated first before we could use the new router options.


apps/dataset-browser/next.config.mjs
sdevalk, line 1: I'm all in favor of using ESM instead of CommonJS. But I prefer the `.js` extension rather than `.mjs`. Is this possible?

apps/dataset-browser/next.config.mjs
```diff
@@ -0,0 +1,17 @@
+/**
+ * @type {import('next').NextConfig}
+ */
+
+import NextIntlPlugin from 'next-intl/plugin';
+import MDXPlugin from '@next/mdx';
+const withNextIntl = NextIntlPlugin('./src/i18n.ts');
```
sdevalk, line 7: Insert a newline, to separate the imports from the rest of the code?

sdevalk (commented):
Nice fix!

apps/dataset-browser/next.config.mjs
barbarah: I don't think so. The documentation of Next.js says: "If you need ECMAScript modules, you can use `next.config.mjs`": https://nextjs.org/docs/pages/api-reference/next-config-js

I needed to move to ESM because, for some reason, the NextIntlPlugin would not import with CommenJs (using require). I will ask the 'next-intl' why this is needed.

sdevalk: Ah, too bad. Perhaps we can revisit this if we - at some point - switch to ESM for our code.

sdevalk (approved):

PR #293, opened 2023-10-23
Custom date range facet for date made

barbarah:
This is just a basic filter with two inputs for date years.

Technically this is the first input where the filter search param type is a number and not an array.


apps/researcher/src/app/[locale]/page.tsx
```diff
@@ -32,17 +33,39 @@ import {ElementType} from 'react';
...
+  {name: 'publishers', searchParamType: 'array'},
+  {name: 'dateCreatedStart', searchParamType: 'number'},
+  {name: 'dateCreatedEnd', searchParamType: 'number'},
+];
+
+const facetComponents: ReadonlyArray<FacetComponent> = [
+  {name: 'owners', Component: FilterSet},
+  {
+    name: 'dateCreated',
+    Component: DateRangeFilter,
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/page.tsx:64]: Would it be possible to align the names of the components? The existing component for facets is called `FilterSet`, the new component for date ranges is called `DateRangeFilter`. Are these both 'filter components'? Would names such as `SelectFilter` (or `MultiSelectFilter`) and `DateRangeFilter` make sense?

apps/researcher/src/app/[locale]/page.tsx
```diff
@@ -32,17 +33,39 @@ import {ElementType} from 'react';
...
+    customProps: {
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/page.tsx:65]: Would it be possible to type these properties, to make clear what the expected properties (names and data types) are? For example by defining a separate `interface`, e.g.

```
interface DateRangeFacet {
  name: string;
  Component: ElementType;
  startDateKey: string;
  endDateKey: string;
}
```

apps/researcher/src/app/[locale]/page.tsx
```diff
@@ -32,17 +33,39 @@ import {ElementType} from 'react';
...
 
-interface FacetProps {
+interface FilterSetting {
   name: keyof SearchResult['filters'];
   searchParamType: SearchParamType;
+}
+
+interface FacetComponent {
+  name: string;
   Component: ElementType;
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/page.tsx:47]: Lowercase `c`, `component`?

apps/researcher/src/app/[locale]/page.tsx
```diff
@@ -32,17 +33,39 @@ import {ElementType} from 'react';
 // Revalidate the page every n seconds
 export const revalidate = 60;
 
-interface FacetProps {
+interface FilterSetting {
   name: keyof SearchResult['filters'];
   searchParamType: SearchParamType;
+}
+
+interface FacetComponent {
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/page.tsx:45]: The interface `FacetComponent` also has a prop called `Component` - this makes the name of the interface somewhat harder to understand for me. Would just the name `Facet` work (i.e. a facet consists of a name and a component)?

apps/researcher/src/messages/nl/messages.json
```diff
@@ -117,8 +117,11 @@
     "deathYearsFilter": "Overlijdensjaren",
     "deathPlacesFilter": "Overlijdensplaatsen",
     "publishersFilter": "Bronhouder objecten",
+    "dateCreatedFilter": "Datum gemaakt",
     "search": "Zoeken",
     "filters": "Filters",
+    "fromYear": "Van jaar",
+    "tillYear": "Tot jaar",
```
sdevalk, line 124: Detail: `Tot en met` (the date search is inclusive)?

apps/researcher/src/app/[locale]/page.tsx
Comment (barbarah) [reply apps/researcher/src/app/[locale]/page.tsx:47]: `ElementType`s need to be uppercase to be valid JSX.

apps/researcher/src/app/[locale]/page.tsx
Comment (sdevalk) [reply apps/researcher/src/app/[locale]/page.tsx]: Perfect!

apps/researcher/src/app/[locale]/page.tsx
Comment (sdevalk) [reply apps/researcher/src/app/[locale]/page.tsx]: Can you this as a comment to the code? Then future readers understand why we're breaking the lowercase convention in interface props.

apps/researcher/src/app/[locale]/page.tsx
```diff
@@ -36,17 +37,43 @@ import {ElementType} from 'react';
...
+  name: string;
   Component: ElementType;
 }
 
-const facets: ReadonlyArray<FacetProps> = [
-  {name: 'owners', searchParamType: 'array', Component: FilterSet},
-  {name: 'types', searchParamType: 'array', Component: FilterSet},
-  {name: 'subjects', searchParamType: 'array', Component: FilterSet},
-  {name: 'publishers', searchParamType: 'array', Component: FilterSet},
+interface DateRangeFacet {
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/page.tsx:59]: Is it possible to extend the `Facet` interface here? Then `DateRangeFacet` can make use of `name` and `Component`?

Comment (barbarah) [reply apps/researcher/src/app/[locale]/page.tsx:59]: Yes, I agree that that is better

apps/researcher/src/app/[locale]/page.tsx
```diff
@@ -36,17 +37,42 @@ import {ElementType} from 'react';
...
+  {name: 'types', searchParamType: 'array'},
+  {name: 'subjects', searchParamType: 'array'},
+  {name: 'publishers', searchParamType: 'array'},
+  {name: 'dateCreatedStart', searchParamType: 'number'},
+  {name: 'dateCreatedEnd', searchParamType: 'number'},
+];
+
+interface Facet {
+  name: string;
+  // ElementType needs to be uppercase to be valid JSX
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/page.tsx:56]: Minor: is it `ElementType` that needs to be uppercase? Or `Component`?

sdevalk (approved):

PR #301, opened 2023-10-25
Bump "@clerk/nextjs" to "4.25.7"

barbarah:
I'm not sure why Dependabot is not updating `@clerk/nextjs`. But it hasn't been updated for 20 days and there have been multiple new versions since then. This new version also includes `membership_count` so I have updated the code to use this property.

I think this update also fixes a bug. I did not see the "add object to list" button on dev.app.colonialcollections.nl after logging in. The button should only be hidden if you are not logged in. I do see it now in Preview (and local development). I'm not sure what is causing this bug. So I will keep an eye on this.

### DIFF
--- a/apps/researcher/src/lib/community.ts
+++ b/apps/researcher/src/lib/community.ts
@@ -8,6 +8,7 @@ export interface Community {
   slug: string;
   imageUrl: string;
   createdAt: number;
+  membersCount?: number;
 }
 
 export interface Membership {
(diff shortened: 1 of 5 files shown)


apps/researcher/src/lib/community.ts
```diff
@@ -8,6 +8,7 @@ export interface Community {
   slug: string;
   imageUrl: string;
   createdAt: number;
+  membersCount?: number;
```
sdevalk, line 11: We also have an enum value called `MembershipCountDesc`. Align the names? E.g. `membershipCount` or `MembersCountDesc` for the enum value? Or something different: `numberOfMembers`?

sdevalk (approved):
👍🏻

PR #315, opened 2023-11-08
Custom edit community form with ORCID iD

barbarah:
This is the "edit community" based on the design: https://gui-prototype.colonialcollections.nl/community.html

This design contains: 

  - A warning banner with ORCID iD
  - A new edit form, including the ORCID iD

I had some discussions with @Doppen about the new edit community form. This form missed a lot of features of the existing Clerk organization profile. And with the new design, the button to the Clerk organization profile was gone. We went through the missing features and decided what to do with them in an ideal situation:

- Edit slug -> move to the new "edit community" form
- Edit image -> move to a new form that opens when clicking on the image
- "Leave Community" button -> Move to page (design needed)
- "Delete Community" button -> Move to page (design needed)
- Edit members (overview, invite, remove, change roles) -> Because this feature contains a lot of logic, use the page Clerk already provides.

In this pull request, I have included the first and last points. The other points would be nice but take some more work, and since Clerk already provides these features, I have added a button, "settings". This will open the settings tab of the Clerk organization profile. If we have time, we could remove the "settings" button and implement the features with our design.

# Technical points

- I needed to downgrade "Zod" because of this issue: https://github.com/colinhacks/zod/issues/2663
- I called the new property `orcid`, but I remember @sdevalk saying he preferred a different name, so other identifiers could also be used. I don't know enough about the scope of this identifier to decide on a name. So other suggestions than `orcid` are welcome.


barbarah:
The end-to-end tests fail because of the reason changes in the filters. I will fix this in a separate pull request.

apps/researcher/src/app/[locale]/communities/[slug]/actions.ts
```diff
@@ -1,22 +1,35 @@
 'use server';
 
-import {joinCommunity, updateDescription} from '@/lib/community';
+import {joinCommunity, updateCommunity} from '@/lib/community';
 import {revalidatePath} from 'next/cache';
 
 interface UpdateDescriptionAndRevalidateProps {
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/communities/[slug]/actions.ts:6]: Or `UpdateCommunityAndRevalidateProps`?

apps/researcher/src/app/[locale]/communities/[slug]/actions.ts
```diff
@@ -1,22 +1,35 @@
...
   communityId: string;
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/communities/[slug]/actions.ts:7]: Or just `id` (since the namespace `updateCommunityAndRevalidate` already makes clear it's about the community)?

apps/researcher/src/app/[locale]/communities/[slug]/actions.ts
```diff
@@ -1,22 +1,35 @@
...
-  communitySlug: string;
+  name: string;
+  slug: string;
   description: string;
+  orcid: string;
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/communities/[slug]/actions.ts:11]: We do not know yet whether we're going to use ORCID IDs. Would it be possible to use a more generic name instead, e.g. `attributionId`?

sdevalk (approved):
Nice!

PR #316, opened 2023-11-10
Refactor the list store, so that there is one store per list

barbarah:
This pull request fixes the "something went wrong." error you sometimes get when navigating between lists (for example, from the object list to the person list).

The problem is that there is one list store; the reset between navigating must be fixed. 

I fixed this by placing each list in its own context. So, each list uses a separate store.

This new structure of the list store is very similar to the setup of the searchable filter store and is based on the Zustand docs: https://docs.pmnd.rs/zustand/guides/typescript#bounded-usestore-hook-for-vanilla-stores

The proposed Typescript solution from Zustand does require a required selector. So not `const store = useListStore()` but `const sortBy = useListStore(s => s.sortBy)`. Using a selector is also preferred because it will limit the react renders.


barbarah:
The end-to-end tests are failing because the back button from the object page also needs the store. But now, the back button does not have this context.

packages/list-store/src/search-params-updater.tsx
```diff
@@ -0,0 +1,61 @@
...
+    newDataNeeded,
+    query,
+    offset,
+    sortBy,
+    selectedFilters,
+    transitionStarted,
+  ]);
+}
+
+// Wrap the hook in an component so can be placed in a context provider
```
sdevalk, line 53: Minor: `so it can`?

sdevalk (approved):
Nice! Just one comment about a comment.

barbarah:
@sdevalk I needed to update this pull request code. The details pages also need the list store to set the correct search parameters when navigating back to the list. With the first version of this pull request, the detail page could not use the list store. With the new changes, every list (datasets, objects, persons, communities) has the next structure.

```
<Layout with a ListProvider> <-- `baseUrl` and `defaultSearch` are set here once

  <Page with a list>
    <ListStoreUpdater /> <-- This will update the list store based on the search params
  </Page with a list>

  <Page with details>
    <ToFilteredListButton />  <-- This button will navigate back using the search params from the store
  </Page with details>

</Layout with a ListProvider>
```

I needed to move the object `page.tsx` and `objects/page.tsx` to the route group `(objects)` so the new layout with the `ListProvider` was only used for the object pages. 

I also fixed a non-related failing end-to-end test.

sdevalk (approved):

PR #325, opened 2023-11-15
Update translations for persons to constituents

barbarah:
Rename 'Persons' to 'Constituents' (English) and 'Betrokkenen' (dutch) 

In code, the entity is still called a person. We also have a person api. Changing this will be some more work.

### DIFF
--- a/apps/researcher/src/messages/nl/messages.json
+++ b/apps/researcher/src/messages/nl/messages.json
@@ -9,7 +9,7 @@
     "filters": "Filters"
   },
   "Constituents": {
-    "title": "{totalConstituents, plural, =0 {0 betrokkenen} =1 {1 betrokken} other {# betrokkenen}}",
+    "title": "{totalConstituents, plural, =0 {0 Personen en Instellingen} =1 {1 Persoon of Instelling} other {# Personen en Instellingen}}",
     "noResults": "Er zijn geen resultaten",
     "fetchError": "Er is een probleem opgetreden bij het ophalen van de resultaten. Laad de pagina opnieuw om het opnieuw te proberen.",
     "filters": "Filters"
@@ -24,7 +24,7 @@
   },
   "Tabs": {
     "objects": "Objecten zoeken",
-    "persons": "Betrokkenen zoeken"
+    "persons": "Personen en Instellingen zoeken"
   },
   "LanguageSelector": {
     "nl": "Nederlands",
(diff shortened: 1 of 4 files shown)


apps/researcher/src/messages/nl/messages.json
```diff
@@ -24,7 +24,7 @@
   },
   "Tabs": {
     "objects": "Objecten zoeken",
-    "persons": "Personen zoeken"
+    "persons": "Personen en Instellingen zoeken"
```
sdevalk, line 27: Lowercase `I`, `instellingen`?

apps/researcher/src/messages/nl/messages.json
```diff
@@ -8,8 +8,8 @@
     "fetchError": "Er is een probleem opgetreden bij het ophalen van de resultaten. Laad de pagina opnieuw om het opnieuw te proberen.",
     "filters": "Filters"
   },
-  "Persons": {
-    "title": "{totalPersons, plural, =0 {0 personen} =1 {1 persoon} other {# personen}}",
+  "Constituents": {
+    "title": "{totalConstituents, plural, =0 {0 Personen en Instellingen} =1 {1 Persoon of Instelling} other {# Personen en Instellingen}}",
```
sdevalk, line 12: Minor:
1. `personen en instellingen`, lowercased?
2. Instead of `totalConstituents` use `totalNumberOfConstituents` or `constituentsTotalCount`, to make clear this is a numeric value?

apps/researcher/src/messages/nl/messages.json
sdevalk: An afterthought: is the section limited to 'instellingen' (i.e. it's not about 'organisaties' in general)?

barbarah: There has been a discussion on Slack about the Dutch translation. I do not know enough about the scope of this section to decide on the correct translation. So I invite you to ask your questions there. The reasoning from Valentin was: On the NMVW collection website, the term used is 'Personen en instellingen'. This is the term Marjolijn prefers. With the reasoning that "I believe Personen and Instellingen would be good to use. 'Betrokkenen' suggests some level of involvement which might not always be there. For instance if an artwork is inspired by another artists, the latter one probably has nothing to do with this piece. So I would prefer a bit more neutral/broader term like Personen and Instellingen."

PR #337, opened 2023-11-24
Make the community page more user friendly

barbarah:
Based on the first feedback here are some updates to the Community page

- The community menu item is removed
- There is a new 'create new community' button
- There is a 'Show only my communities' toggle
- I have added the missing "Order by members" options

This is based on this design: https://gui-prototype.colonialcollections.nl/all-community.html

### DIFF
--- a/apps/researcher/src/app/[locale]/communities/page.tsx
+++ b/apps/researcher/src/app/[locale]/communities/page.tsx
@@ -9,6 +9,8 @@ import {
   SearchField,
   OrderSelector,
 } from '@colonial-collections/ui/list';
+import {AddCommunityButton} from './buttons';
+import {MyCommunityToggle} from './my-community-toggle';
 
 // 1 day = 60*60*24 = 86400
 export const revalidate = 86400;
@@ -21,6 +23,7 @@ interface Props {
     query?: string;
     sortBy?: SortBy;
     offset?: number;
+    onlyMyCommunities?: number;
   };
 }
 
@@ -38,6 +41,7 @@ export default async function CommunitiesPage({
       query,
       sortBy,
       offset,
+      onlyMyCommunities: !!searchParams.onlyMyCommunities,
     });
   } catch (err) {
     return <ErrorMessage error={t('error')} />;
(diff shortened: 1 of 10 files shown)


apps/researcher/src/app/[locale]/communities/page.tsx
```diff
@@ -21,6 +23,7 @@ interface Props {
     query?: string;
     sortBy?: SortBy;
     offset?: number;
+    onlyMyCommunities?: number;
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/communities/page.tsx:26]: The name `onlyMyCommunities` suggests this is a `boolean`, not a `number`. Is there a way to either rename the property or to change the type? If not, add a comment?

apps/researcher/src/messages/en/messages.json
```diff
@@ -161,7 +162,9 @@
     "membershipCount": "{count, plural, =0 {0 members} =1 {1 member} other {# members}}",
     "objectListCount": "{count, plural, =0 {0 lists} =1 {1 list} other {# lists}}",
     "objectListCountError": "Error",
-    "searchPlaceholder": "Search for community"
+    "searchPlaceholder": "Search for community",
+    "addCommunity": "Create a community",
+    "toggleMyCommunities": "Show only my communities"
```
sdevalk, line 167: Or rename `toggleMyCommunities` to `showMyCommunities` (to make its meaning clearer)?

apps/researcher/src/app/[locale]/communities/page.tsx
Comment (barbarah) [reply apps/researcher/src/app/[locale]/communities/page.tsx:26]: `number` is wrong. It is either the string 'true' or undefined, so I have renamed it to `onlyMyCommunities?: 'true';`

barbarah:
I have fixed the comments. I have also decided to create a separate `getMyCommunities` function. Instead of one `getCommunities` function with a `onlyMyCommunities` property.

 By separating the logic I can add `noStore()` to  `getMyCommunities` (where the response can change depending on the logged-in status) without disabling the cache for `getCommunities`.

sdevalk (approved):

PR #344, opened 2023-11-28
feat: update nanopubs code according to new model

sdevalk:
This (backward incompatible!) PR updates the code of the `enricher` package according to the new data model of our nanopublications. The model is not finished yet, but the code should be more or less stable by now (although still unoptimized).

The most significant changes:

1. The new `creator` class should be used for creating enrichments.
2. You can now store the language of the enrichment, in the `inLanguage` prop
3. You can now store the name of the user of the enrichment (`creator` > `name`)
4. You can now indicate the field that the enrichment is about (e.g. about the `name` or `description`). The supported fields are [here](https://github.com/colonial-heritage/colonial-collections/blob/feat-nanopubs-model/packages/enricher/src/definitions.ts#L10-L20)
5. The `fetcher` allows you to fetch all the enrichments about a resource (in our case: a heritage object). In the frontend we'll need to filter the enrichments, so that we can display the enrichments at the right positions in the UI (e.g. the enrichments about the `name` should be presented in a different place than the enrichments about the `description`). If this  filtering is cumbersome, the `fetcher` could do additional work (see [here](https://github.com/colonial-heritage/colonial-collections/blob/feat-nanopubs-model/packages/enricher/src/fetcher.ts#L135))?


packages/enricher/src/helpers.ts
```diff
@@ -0,0 +1,86 @@
...
+import type {Resource} from 'rdf-object';
+
+export function fromAboutTypeToClass(additionalType: AdditionalType) {
+  let className;
+  if (additionalType === AdditionalType.Description) {
+    className = 'Description';
+  } else if (additionalType === AdditionalType.Name) {
+    className = 'Name';
+  } else {
+    throw new TypeError(`Unknown type: "${additionalType}"`);
```
barbarah, line 17: Only `Name` and `Description` are supported here, but there are more additional types, are there rest unsupported?

packages/enricher/src/definitions.ts
```diff
@@ -1,14 +1,62 @@
...
+export enum AdditionalType {
+  Creator = 'creator',
+  DateCreated = 'dateCreated',
+  Description = 'description',
+  Inscription = 'inscription',
+  Material = 'material',
+  Name = 'name',
+  Subject = 'subject',
+  Technique = 'technique',
+  Types = 'types',
```
barbarah, line 19: Why is 'Types' plural but all other fields singular?

sdevalk: Wel-spotted! Fixed.

packages/enricher/src/helpers.ts
sdevalk: I've just updated the code - the other types are now also being supported.

barbarah (approved):
Nice! I have approved this pull request. But because of the braking changes it should not be merged without the frontend changes.

PR #352, opened 2023-12-05
Title and language to metadata

barbarah:
Based on this design: https://gui-prototype.colonialcollections.nl/object.html

image

1. Under the title, there are max 3 alternative object names
2. Under Metadata, the title is shown the same as all other metadata 
3. The label of the 'Add' button is longer 
4. If there is a language code, show the language in the metadata


apps/researcher/src/app/[locale]/(objects)/objects/[id]/page.tsx
```diff
@@ -89,6 +91,9 @@ export default async function Details({params}: Props) {
 
   const enrichments = await fetcher.getById(id);
   useObject.setState({objectId: object.id, locale, enrichments});
+  const enrichmentNames = enrichments?.filter(
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/(objects)/objects/[id]/page.tsx:94]: To be sure: what are 'enrichment names'? For example, are these enrichments about the name of an object? If so, rename to e.g. `enrichmentsAboutName` or `nameEnrichments`?

PR #363, opened 2023-12-09
feat: improve datasets api

sdevalk:
This PR improves the backend API of the `dataset-browser`. The primary changes:

1. The API now uses SPARQL for fetching the metadata of datasets (instead of Elasticsearch)
2. The structure of the API now follows the API of the `researcher`
3. The API no longer supports the dataset props `spatialCoverages` and `genres` - these props are not being used by datasets (on production). We can add them again when needed
4. The dataset props `publisher` and `license` have become optional (instead of required). This is a small breaking change.


barbarah (commented):
The end-to-end tests fail for the datasets. The reason is that the filters with 0-values in the search results are not returned any more. So the test 'filters all categories together (query, license and publisher)' first searches for 'My query', and it will not return any licenses and publishers.

The fix is easy. Change the search text from 'My query' to 'dataset', and you will get filters in the search results.

apps/dataset-browser/src/app/[locale]/datasets/[id]/page.tsx
```diff
@@ -59,18 +59,14 @@ export default async function Details({params}: Props) {
     {
       name: t('metadata.license'),
       value: (
-        <a href={dataset.license.id} target="_blank" rel="noreferrer">
-          {dataset.license.name}
+        <a href={dataset.license?.id} target="_blank" rel="noreferrer">
```
Comment (barbarah) [inline apps/dataset-browser/src/app/[locale]/datasets/[id]/page.tsx:62]: If no `license` exists, there should not be an empty `a' tag.

apps/dataset-browser/src/app/[locale]/datasets/[id]/page.tsx
```diff
@@ -59,18 +59,14 @@ export default async function Details({params}: Props) {
...
       ),
     },
     {
       name: t('metadata.publisher'),
-      value: (
-        <a href={dataset.publisher.id} target="_blank" rel="noreferrer">
-          {dataset.publisher.name}
-        </a>
-      ),
+      value: dataset.publisher?.name,
```
Comment (barbarah) [inline apps/dataset-browser/src/app/[locale]/datasets/[id]/page.tsx:69]: Just a note: I think users would like a link to the publisher, but we can move this to a feature request when users ask. If the data is absent, I agree with removing the link tag.

barbarah:
This pull request will probably also fix: https://github.com/colonial-heritage/colonial-collections/pull/340

Since the file that shows typescript errors is removed.

apps/dataset-browser/src/app/[locale]/datasets/[id]/page.tsx
Comment (sdevalk) [reply apps/dataset-browser/src/app/[locale]/datasets/[id]/page.tsx:69]: The previous code assumed `dataset.publisher.id` pointed to a website, but it's actually an (linked data) identifier that doesn't have to resolve to a webpage that's useful to users. I agree - if users want to be able to go to a website, we should created a feature request.

apps/dataset-browser/src/app/[locale]/dataset-card.tsx
```diff
@@ -1,6 +1,6 @@
 import {Link} from '@/navigation';
 import {useTranslations} from 'next-intl';
-import {Dataset} from '@/lib/datasets';
+import {Dataset} from '@/lib/api/datasets';
 import {Badge} from '@colonial-collections/ui';
 import {
   GlobeEuropeAfricaIcon,
```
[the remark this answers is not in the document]
Comment (sdevalk) [reply apps/dataset-browser/src/app/[locale]/dataset-card.tsx:6]: That makes sense, thanks! I'll remove the definitions then.

barbarah (approved):

PR #368, opened 2023-12-12
feat: add location of creation to object details

sdevalk:
This PR adds the location of creation to an heritage object [1]. The location is currently only visible on the search page, in a facet, not on the detail page of an object. The data structure looks like this:

```json
{
  "id": "http://...",
  "locationCreated": {
    "id": "http://...",
    "name": "Jakarta",
    "isPartOf": { // Optional - may not exist (e.g. if the location above is a country)
      "id": "http://...",
      "name": "Indonesia",
    }
  }
}
```

This PR is rather large: it also updates the unit tests and brings the backend APIs in the `researcher` more in line with the backend API of the `dataset-browser`. All changes should be backward-compatible, though.

[1] https://github.com/orgs/colonial-heritage/projects/1/views/1?pane=issue&itemId=43426571

### DIFF
--- a/apps/researcher/package.json
+++ b/apps/researcher/package.json
@@ -36,6 +36,7 @@
     "@hookform/resolvers": "3.3.2",
     "@next/mdx": "14.0.1",
     "classnames": "2.3.2",
+    "defu": "6.1.3",
     "fetch-sparql-endpoint": "4.1.0",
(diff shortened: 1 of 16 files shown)


barbarah (approved):
Looks good. We probably need to add `locationCreated` to the details page somewhere.

PR #369, opened 2023-12-14
fix: IDs of prov events

sdevalk:
This PR fixes the IDs that were used in the provenance events: the code previously used IDs of the parent events of provenance events, not the IDs of the actual events. This made the coherence of provenance events strange.


barbarah (approved):
👍

PR #370, opened 2023-12-14
Add provenance data table

barbarah:
This pull request will add the provenance data table. 

It's built in an expandable way, so we can easily add the next provenance blocks (timeline and map).

All events have a label starting with the first letter of 'provenance' in the language selected plus a number. You can click on a provenance label, this will set the event to selected in the `SelectedEventProvider`. Since the other provenance blocks aren't present, selecting an event will not do anything for the user.

The events are sorted based on their `startsAfter` and `endsBefore` properties. (Maybe this sorting needs to be moved to the backend?). And grouped by date range.

If the event has a description, you can click a slideout button, so you can read the description.


sdevalk:
Regarding:

> The events are sorted based on their `startsAfter` and `endsBefore` properties. (Maybe this sorting needs to be moved to the backend?). And grouped by date range.

Yes, we could consider this - there's an old comment in the code about sorting [1]. I've never implemented this, though, because sorting the events is rather complicated and because so far we haven't seen real provenance data; it could be that the real data cannot make use of the `startsAfter` and `endsBefore` props (in which case we'll have to rethink the sorting).

At this stage I'm happy with the current approach: doing the sorting in the frontend (= this PR). We can revisit this later, if needed.

[1] https://github.com/colonial-heritage/colonial-collections/blob/main/apps/researcher/src/lib/api/objects/provenance-events-fetcher.ts#L369

apps/researcher/src/app/[locale]/(objects)/objects/[id]/(provenance)/group-events.test.ts
```diff
@@ -0,0 +1,41 @@
+import {describe, expect} from '@jest/globals';
+import {groupByDateRange} from './group-events';
+import {LabeledProvenanceEvent} from './definitions';
+const locale = 'en';
+describe('groupByDateRange', () => {
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/(objects)/objects/[id]/(provenance)/group-events.test.ts:5]: For readability, put a newline before and after the `const locale`?

sdevalk (approved):
Perfect! Just one comment.

PR #373, opened 2023-12-15
feat: add license of image to API

sdevalk:
This PR adds a license to the image of an object. The license is optional - data providers may not have it.


barbarah (approved):
It's working as expected. I will create a pull request for the front end soon.

PR #374, opened 2023-12-15
Add location created to object page

barbarah:


apps/researcher/src/app/[locale]/(objects)/objects/[id]/page.tsx
```diff
@@ -206,6 +206,12 @@ export default async function Details({params}: Props) {
             >
               {object.description}
             </Metadata>
+            <Metadata
+              translationKey="locationCreated"
+              enrichmentType={AdditionalType.LocationCreated}
+            >
+              {object.locationCreated?.name}
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/(objects)/objects/[id]/page.tsx:213]: This show the actual location, right? For consideration: we could also show the country of which it is a part (`isPartOf`), depending on the UI wishes (e.g. `Jakarta (Indonesia)`).

sdevalk (approved):

apps/researcher/src/app/[locale]/(objects)/objects/[id]/page.tsx
Comment (barbarah) [reply apps/researcher/src/app/[locale]/(objects)/objects/[id]/page.tsx]: Good point, I have added `isPartOf`

PR #389, opened 2023-12-20
Show correct facet background for the old design of the Dataset Browser

barbarah:
The dataset browser has not yet moved to the new colours, but it shares components with the research app.

For most components, this is not a big problem, but the facets aren't readable. So I have added a temporary variant property. If we decide on more changes between the apps we could think about adding a theme.

Old:
image

With the changes is this pull request:
image


sdevalk (approved):
👍🏻

PR #390, opened 2023-12-21
API: maintenance

sdevalk:
This is a maintenance PR:

1. It updates the environment variables, replacing the old hostname of our TriplyDB instance with the new one
2. It updates the tests that referred to the old hostname
3. It removes the 'owner' property from a heritage object and corresponding facet from the search results: the owner is not used in the production data. We can add it later when needed.


barbarah (approved):
Looks good. I'm happy with the removal of 'owner'. Now I wont accidentally pick 'owner' instead of 'publisher' anymore.

PR #394, opened 2023-12-21
fix: filter with AND instead of OR

sdevalk:
This PR fixes [1] by changing the filtering behaviour of our Elasticsearch client: it now filters by 'and' instead of 'or'. This causes the search results to be narrowed down significantly.

We should test whether this indeed is the desired behaviour. For example, the effect is that users will see less facet values to choose from. This could hamper their search experience.

[1] https://github.com/orgs/colonial-heritage/projects/1/views/1?pane=issue&itemId=48168750


barbarah (approved):
Nice!

PR #395, opened 2023-12-22
feat: use SPARQL for fetching the search results

sdevalk:
This PR updates the search API of the Research App. We now use Elasticsearch for fetching search results. But SPARQL gives much greater power. This PR replaces ES with SPARQL. This is a backward-compatible change and doesn't (or shouldn't) impact the frontend.


barbarah (approved):
Looks good. All id's will be urls after this change? This should not be a problem. But we should tests the filtering after this change just to be sure.

sdevalk:
> Looks good. All id's will be urls after this change? This should not be a problem. But we should tests the filtering after this change just to be sure.

Good question. The IDs in the facets will continue to be literals/names, as is currently the case: these literals still come from Elasticsearch. But the IDs in the details of heritage objects (i.e. the search results) will become URLs.

PR #397, opened 2023-12-22
Community bugs

barbarah:
Fix 3 bugs on the community page.

## 1. [Only set community active when signed in](https://github.com/colonial-heritage/colonial-collections/commit/bf7a828aa691b7ea95cc404b7367db826b0e1db2)

The app tries to set the community to active when not signed in. This will result in an error (and failing end-to-end tests)

## 2. [Load async component with Suspense](https://github.com/colonial-heritage/colonial-collections/commit/bb97d0f633604d4c55f71578e53930cf985585cb)

The `LocalizedMarkdown` component is an async component. This does not load well in client components. The old workaround does not work anymore (probably since a package update; I'm unsure when). When you open the community form with the `LocalizedMarkdown,` you will get an error. The new solution is added a `Suspense` wrapper.

## 3. [Only community name is required](https://github.com/colonial-heritage/colonial-collections/pull/397/commits/258b63e280969d4de75f2e29bad041ec4a40626a)

 In the community form, all fields had a "*". But only the name is required, so I removed the others.


sdevalk (approved):
👍🏻

PR #399, opened 2023-12-22
fix: allow for multiple locations of creation

sdevalk:
This PR fixes an issue in the 'fetch' API. The API allowed for just one location of creation of an object. But apparently - according to the production data - an object can have multiple locations of creation.


apps/researcher/src/app/[locale]/objects/[id]/page.tsx
```diff
@@ -218,9 +218,13 @@ export default async function Details({params}: Props) {
                 translationKey="locationCreated"
```
Comment (barbarah) [inline apps/researcher/src/app/[locale]/objects/[id]/page.tsx:218]: Can you also rename the translation key to 'locationsCreated' (also in the messages files)?

barbarah (approved):
Nice, I do have one small comment

PR #414, opened 2024-01-11
Markup text pages

Doppen:
Text pages had no margin and alignments. Added markup.

### DIFF
--- a/packages/ui/localized-markdown.tsx
+++ b/packages/ui/localized-markdown.tsx
@@ -20,7 +20,6 @@ export async function LocalizedMarkdown({name, contentPath, textSize}: Props) {
     'prose-h2:text-xl prose-h2:mb-1 prose-h2:mt-10',
     'prose-h3:text-lg prose-h3:mb-2',
     'prose-a:text-sky-700 prose-a:no-underline',
-    'mx-auto mt-16 px-4 sm:px-10 mb-16',
     {
       'text-sm': textSize === 'small',
     }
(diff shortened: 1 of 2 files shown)


packages/ui/localized-markdown.tsx
```diff
@@ -20,6 +20,7 @@ export async function LocalizedMarkdown({name, contentPath, textSize}: Props) {
     'prose-h2:text-xl prose-h2:mb-1 prose-h2:mt-10',
     'prose-h3:text-lg prose-h3:mb-2',
     'prose-a:text-sky-700 prose-a:no-underline',
+    'mx-auto mt-16 px-4 sm:px-10 mb-16',
```
barbarah, line 23: This component is used on other pages than just the information pages. So adding padding will effect more then just the information pages.

image

A better solution would be adding the padding to `apps/researcher/src/app/[locale]/(information)/layout.tsx`

Something like (not tested): 

```
import BaseLayout from '@/components/base-layout';
import {ReactNode} from 'react';

export default function Layout({children}: {children: ReactNode}) {
  return (
    <body>
      <BaseLayout wrapperClassName="mx-auto mt-16 px-4 sm:px-10 mb-16">
        {children}
      </BaseLayout>
    </body>
  );
}
```

Doppen: Ah, indeed. Tnx

barbarah (approved):

PR #422, opened 2024-01-23
fix: sort by date created, no longer by relevance

sdevalk:
This PR:

1. Removes the 'sort by relevance' option from the 'heritage objects' API
2. Adds the 'sort by date made' option to the 'heritage objects' API

Beware: this PR only changes the API - the parts in the frontend using the API will break. @barbarah Could you determine how you want to integrate this API change in the frontend?


barbarah (approved):
I have added one commit to fix the frontend. Please check if you agree with the changes.

sdevalk:
> I have added one commit to fix the frontend. Please check if you agree with the changes.

Perfect, thanks! I've made a small change to the messages in the locale files, so that the words are in sync with the words we already use (e.g. 'date made', not 'date created').

PR #423, opened 2024-01-23
Wrap `SignedIn` component in client component

barbarah:
There is still a bug with the Clerk code in our app. Sometimes the session seems broken. The backend clerk logic will not see the user as signed in. Only removing the cookies helps. Refreshing or sign-in/sign-out will not fix this bug. I cannot find/fix the bug, and asking Discord did not help. So I'm now working around the bug.

After analyzing a broken session, I could see the new custom Protect component I built works correctly. The clerk `SignedIn` in components does not work correctly within server components. So, I have wrapped the `SignedIn` component so it will always render as a client component. I think this will work because it also works for the Protect component.

Again: this is hard to test because I cannot reproduce the bug.


apps/researcher/src/lib/community/signed-in.tsx
```diff
@@ -0,0 +1,14 @@
+'use client';
+
+import {SignedIn as ClerkSignedIn} from '@clerk/nextjs';
+import {ReactNode} from 'react';
+
+/*
+Use this component to wrap content that should only be visible to signed in users.
+Do not use the <SingIn /> component from @clerk/nextjs directly,
```
sdevalk, line 8: Minor: `do not use the <SignedIn /> component...`? Although I do like to sing in :) 🎶🎶

apps/researcher/src/lib/community/signed-in.tsx
```diff
@@ -0,0 +1,14 @@
...
+as it will sometimes not work in server components.
+By wrapping it in this client component, we can ensure that it will always work.
```
sdevalk, line 10: Minor: `By wrapping it in this client component it always works.` (active voice)?

sdevalk (approved):
I hope this resolves the issue :)

I've added two minor suggestions.

PR #425, opened 2024-01-23
Remove tabs from object search results page

barbarah:
Ticket https://github.com/colonial-heritage/sprints/issues/360

The 'constituents' page is not helpful with the current implementation/data. For now, remove the tabs. The page does not need to be removed. So we can still navigate to the page if needed.


apps/researcher/src/app/[locale]/objects/page.tsx
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/objects/page.tsx:1]: There's a GitHub warning about an undefined `Tabs`.

sdevalk (approved):
Just one comment

PR #430, opened 2024-01-25
fix: dates

sdevalk:
This PR fixes some issues that we have with dates of objects and provenance events:

1. Negative dates (e.g. `-1714`) are currently changed to positive dates (e.g. `1714`)
2. Dates with only years (e.g. `1948`) are currently changed to January 1 of these years (e.g `1948-01-01`), which could suggest that objects weren't made or provenance events didn't took place in that year.

The PR addresses these issues by introducing a new package, EDTF. This package parses the start date and end date of an object and its provenance events. It returns a string notation that respects the negative dates, sets a start date such as `1947` to `1947-01-01` and an end date such as `1948` to `1948-12-31`, to make clear that the object was created somewhere in that year or that a provenance event took place somewhere in that year.

In its slipstream this PR also fixes an issue where the API attempts to fetch results from its SPARQL endpoint even though Elasticsearch already made clear that there are no results.

### DIFF
--- a/apps/researcher/src/lib/api/objects/rdf-helpers.ts
+++ b/apps/researcher/src/lib/api/objects/rdf-helpers.ts
@@ -1,5 +1,6 @@
 import {Agent, Dataset, Image, License, Place, TimeSpan} from '../definitions';
 import {getProperty, getPropertyValue} from '../rdf-helpers';
+import edtf from 'edtf';
 import type {Resource} from 'rdf-object';
 
 function createThingFromProperty<T>(thingResource: Resource) {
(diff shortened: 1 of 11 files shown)


apps/researcher/src/lib/api/objects/rdf-helpers.ts
```diff
@@ -104,27 +105,29 @@ export function createPlaces(resource: Resource, propertyName: string) {
...
 
   const rawEndDate = getPropertyValue(timeSpanResource, 'ex:endDate');
-  const endDate = fromStringToDate(rawEndDate);
+  const endDate = fromStringToEdtf(rawEndDate);
 
   const timeSpan: TimeSpan = {
     id: timeSpanResource.value,
-    startDate,
-    endDate,
+    startDate: startDate !== undefined ? new Date(startDate.min) : undefined, // E.g. 1881 = 1881-01-01
```
barbarah, line 129: Can we keep it an edtf date? It would be nice to show the date in the frontend as given by the museum, so:

1881 = 1881 
1881-01-01 = 1881-01-01

But this conversion removes the possibility to render the date as given by the museum.

barbarah: Decided offline: We will keep it as a date, and we will have a separate conversation about the formatting of dates

barbarah (approved):

apps/researcher/src/lib/api/objects/rdf-helpers.ts
sdevalk: Small addition: also find out whether there is a need for the API to keep working with Date objects. If we need to do date-time processing (e.g. sorting), dates-as-strings will not suffice. But if there is no necessity, now or in the future, the API might as well simply expose the dates as strings.

PR #432, opened 2024-01-25
Branding less blue object search page

Doppen:
Added more white to the faceted search page.


Doppen:
@barbarah I see some errors with `data-testid`. is that something I can fix?

barbarah:
There are some problems with the test servers, which can hinder the end-to-end tests. For now, you can ignore it, but I will check again tomorrow.

barbarah (approved):

PR #441, opened 2024-01-30
Add locale to api requests

barbarah:


apps/researcher/src/navigation.ts
```diff
@@ -1,6 +1,11 @@
 import {createSharedPathnamesNavigation} from 'next-intl/navigation';
 
-export const locales = ['en', 'nl'] as const;
+export enum LocaleEnum {
```
sdevalk, line 3: For consideration: somehow I would expect these locale values to be in a file with a different name - I probably wouldn't look in `navigation.ts` for these values but rather in, say, a `settings.ts` or `definitions.ts` file.

sdevalk (approved):
Nice! I've added one suggestion (feel free to ignore if it doesn't make sense).

apps/researcher/src/navigation.ts
barbarah: I agree, I will move it to `definitions.ts`

PR #442, opened 2024-01-31
Display BCE dates correctly

barbarah:


apps/researcher/src/app/[locale]/objects/[id]/page.tsx
```diff
@@ -29,7 +30,7 @@ export default async function Details({params}: Props) {
   const object = await heritageObjects.getById({id});
   const locale = useLocale();
   const t = await getTranslations('ObjectDetails');
-  const format = await getFormatter();
+  const formatDate = await getDateFormatter();
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/objects/[id]/page.tsx:33]: Is it possible to align the name of the variable with the name of the function? I would expect a `getDateFormatter()` to return a `dateFormatter`, not a `formatDate`.

apps/researcher/src/lib/get-format-date.ts
sdevalk, line 1: Perhaps rename the filename to `date-formatter.ts`?

apps/researcher/src/lib/get-format-date.ts
```diff
@@ -0,0 +1,31 @@
+'use server';
+
+import {getLocale, getTranslations} from 'next-intl/server';
+
+export async function getDateFormatter() {
```
sdevalk, line 5: If I understand this function correctly it does two things:

1. Create an instance of a date formatter (`formatter`)
2. Format a date (`formatDate`)

The function name, to me, suggests that it returns the instance of (1), but it actually returns a function that does (2)? If so, is it possible to rename the function to make this clearer? For example: `formatDate(date: Date)`?

apps/researcher/src/lib/get-format-date.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  const formatDate = (date?: Date) => {
+    if (!date) {
+      return undefined;
+    }
+
+    const isBCE = date.getFullYear() < 0;
+
+    if (isBCE) {
+      // For dates before 0, we don't want to show the default 'BC' era label.
```
sdevalk, line 22: Perhaps briefly explain why we don't want that, so as to document the choices?

apps/researcher/src/app/[locale]/objects/[id]/page.tsx
Comment (barbarah) [reply apps/researcher/src/app/[locale]/objects/[id]/page.tsx]: This code is removed.

apps/researcher/src/lib/date-formatter.test.ts
```diff
@@ -0,0 +1,38 @@
...
+  },
+});
+
+jest.mock('next-intl/server', () => ({
+  ...jest.requireActual('next-intl/server'),
+  getTranslations: async () => new Promise(resolve => resolve(translate)),
+  getLocale: async () => new Promise(resolve => resolve('en')),
+}));
+
+describe('getDateFormatter', () => {
```
sdevalk, line 18: `formDate`?

apps/researcher/src/lib/date-formatter.ts
```diff
@@ -0,0 +1,34 @@
...
+import {getLocale, getTranslations} from 'next-intl/server';
+
+export async function formatDate(date?: Date) {
+  const t = await getTranslations('FormatDate');
+  const locale = await getLocale();
+  const formatter = new Intl.DateTimeFormat(locale, {
+    year: 'numeric',
+    month: 'short',
+    day: 'numeric',
+  });
```
sdevalk, line 12: Out of curiosity: are there performance considerations for getting translations, locales and creating new formatter instances every time `formatDate` is called? Is it necessary to pre-fetch or cache these or doesn't it have a noticeable perf impact?

barbarah: Yes, there can be performance considerations when fetching translations and locales and creating new formatter instances every time `formatDate` is called. Creating a new `Intl.DateTimeFormat` instance is also not free, although it's likely less expensive than fetching translations or locales.

Taking this into consideration, I'm going back to the old design. I will call the function `getFormatDate`.

sdevalk: Yes, that makes sense!

A final consideration: if we, at some point, want to call `getFormatDate()` in different places/components, every call will load the translations and the locale and create a datetime formatter. Perhaps we should then refactor and put these objects in some global place (e.g. a [registry](https://martinfowler.com/eaaCatalog/registry.html)).

sdevalk (approved):

PR #444, opened 2024-01-31
Search object page

barbarah:
Ticket: https://github.com/colonial-heritage/sprints/issues/367

- Add **Search Objects** to top navigation (see https://gui-prototype.colonialcollections.nl/v2/search-init.html)
- If you click on this item, go to the faceted search page. Only a text search is available (see https://gui-prototype.colonialcollections.nl/v2/search-init.html)
- After the text search, the facets will appear. (https://gui-prototype.colonialcollections.nl/v2/search.html)
- Only after removing all facets/search will you go back to the initial page without facets (previously, it was after removing the search, even if there are still facets selected)
- If you navigate to 'search object' from a different page, the user will see the last search he made

### DIFF
--- a/apps/researcher/src/app/[locale]/objects/initial-search-field.tsx
+++ b/apps/researcher/src/app/[locale]/objects/initial-search-field.tsx
@@ -1,0 +1,17 @@
+'use client';
+
+import {SearchFieldWithLabel} from '@colonial-collections/ui/list';
+import {useRouter} from '@/navigation';
+
+export function InitialSearchField() {
+  const router = useRouter();
+
+  const navigateOnSearch = (query: string) => {
+    if (query) {
+      const urlSearchParams = new URLSearchParams({query});
+      router.replace(`/objects?${urlSearchParams}`);
+    }
+  };
+
+  return <SearchFieldWithLabel onSearch={navigateOnSearch} />;
+}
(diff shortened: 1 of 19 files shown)


apps/researcher/src/app/[locale]/objects/initial-search-field.tsx
```diff
@@ -0,0 +1,17 @@
+'use client';
+
+import {SearchFieldWithLabel} from '@colonial-collections/ui/list';
+import {useRouter} from '@/navigation';
+
+export function InitialSearchField() {
+  const router = useRouter();
+
+  const navigateOnSearch = (query: string) => {
+    if (query) {
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/objects/initial-search-field.tsx:10]: To be sure: does this check rule out a `query` such as `0`? If so, should the check be: `if (query.length > 0)`?

apps/researcher/src/lib/last-search.ts
```diff
@@ -0,0 +1,14 @@
...
+export const useLastSearch = create(() => ({lists: new Map()}));
+
+export function saveLastSearch(baseUrl: string, urlWithSearchParams: string) {
+  useLastSearch.setState(prev => ({
+    lists: new Map(prev.lists).set(baseUrl, urlWithSearchParams),
+  }));
+}
+
+export function getLastSearch(baseUrl: string) {
+  console.log(useLastSearch.getState().lists.keys());
```
sdevalk, line 12: Remove this line?

sdevalk (approved):
Nice! I've added two minor comments.

apps/researcher/src/lib/last-search.ts
barbarah: Whoops..

PR #445, opened 2024-02-01
Display licence Id if name is empty

barbarah:
A minor bug I noticed during the daily when this object was opened: https://app.colonialcollections.nl/nl/objects/https%3A%2F%2Fn2t%252Enet%2Fark%3A%2F27023%2F233a902404f1e8defbbd988cb9aec542

When an object has a licence but no licence text, the licence is not visible.

image

Show the ID if there is no license text. The example object will now be displayed as:

image


sdevalk (approved):
Well-spotted!

PR #451, opened 2024-02-02
feat: locale aware publisher and prov types

sdevalk:
This PR makes the names of publishers/organizations and the names of provenance types locale-aware (similar to the previous PR, in which the names of locations were made locale-aware).

Beware: do not merge this PR just yet - the Elasticsearch indexes on production do not yet have the locale names.


barbarah (approved):
Looks good!

PR #459, opened 2024-02-06
Refactor date formatting so date ranges can accept BCE dates

barbarah:
**The problem**
The provenance data table shows date ranges in the subheaders. The old date range logic had some problems.

1. The date range logic uses `formatRange` from `Intl.DateTimeFormat`. `formatRange` does not accept the `era` param, so it is impossible to show BCE dates correctly.
2. The function showed the same results for (1) A range with the same `startDate` and `endDate`, (2) a range with a missing `startDate` and (3) a range with a missing `endDate`. It is relevant information for the user to know if the start or end date is missing, so we must show this differently. (In the timeline, dates with missing start or end dates will also be displayed differently, but not in this pull request)
3. There are now two functions formatting dates; I want them combined into one formatter.

**The solution**
I wanted one date formatter with two functions: `formatDate` and `formatDateRange`. Unfortunately, this took more code than expected. The formatter uses functions of `next-intl`. But `next-intl` uses different code for client or server code. The formatter is used in both client and server code, so I needed to create two versions. 

I designed the code similarly to the `next-intl` code. A `getFormatter` is used for server code, and `useFormatter` is used for client code. Both will run a `createFormatter` function with the correct translator and formatter. This setup is very similar to the `next-intl` code.

In the provenance code, I moved the `groupByDateRange` to the provenance store, so it only runs once, instead of twice, for the data table and timeline.


apps/researcher/src/app/[locale]/objects/[id]/(provenance)/group-events.test.ts
```diff
@@ -2,7 +2,20 @@ import {describe, expect} from '@jest/globals';
 import {groupByDateRange} from './group-events';
 import {LabeledProvenanceEvent} from './definitions';
 
-const locale = 'en';
+// Is simple `formatDateRange` mock that returns a string representation of a date range.
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/objects/[id]/(provenance)/group-events.test.ts:5]: Minor: instead of `Is` --> `A`?

apps/researcher/src/app/[locale]/objects/[id]/(provenance)/group-events.test.ts
```diff
@@ -2,7 +2,20 @@ import {describe, expect} from '@jest/globals';
...
+const formatDateRange = ({
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/objects/[id]/(provenance)/group-events.test.ts:6]: For consistency with other parts of the codebase, use `function formatDateRange()`?

apps/researcher/src/lib/date-formatter/actions.ts
```diff
@@ -0,0 +1,11 @@
+import {getFormatter, getTranslations} from 'next-intl/server';
+import {createFormatter} from './create-formatter';
+
+export async function getFormatDate(): Promise<
+  ReturnType<typeof createFormatter>
+> {
+  const t = await getTranslations('FormatDate');
+  const format = await getFormatter();
```
sdevalk, line 8: Intuitively I would expect a `getFormatter()` to return a `formatter` (not a `format`). To be sure: what does it return?

apps/researcher/src/lib/date-formatter/create-formatter.test.ts
```diff
@@ -0,0 +1,90 @@
...
+  });
+
+  it('returns undefined if date is not provided', () => {
+    const result = formatDate({date: undefined, t, formatDateTime});
+    expect(result).toBeUndefined();
+    expect(t).not.toHaveBeenCalled();
+    expect(formatDateTime).not.toHaveBeenCalled();
+  });
+
+  it('formats the date with the date formatter for dates after 0', () => {
```
sdevalk, line 25: The tests on line 25 and 31 cover different use cases:

1. Format with the date formatter for dates after 0
2. Format dates using translations for dates before 0

This does not include use cases such as:

1. Format with the date formatter for dates before 0
2. Format dates using translations for dates after 0

To be sure: is this correct?

apps/researcher/src/lib/date-formatter/hooks.tsx
sdevalk, line 1: To be sure: does the extension need to be `.tsx` instead of `.ts`?

sdevalk (commented):
Nice work! I've added some comments and questions.

apps/researcher/src/lib/date-formatter/actions.ts
barbarah: This is the formatter of [`next-intl`](https://next-intl-docs.vercel.app/docs/usage/dates-times). I followed the examples of `next-intl`, and they call 'it' `format`. This is an object with different formatting functions (`dateTime`, `number`, `relativeTime` and `list`).

I think you are right, we can call this `formatter`.

barbarah: Since we create something similar but with custom format functions, I will rename our functions to `useDateFormatter` and `getDateFormatter`.

I have also added a `useMemo` to `useDateFormatter`, inspired by [next-intl](https://github.com/amannn/next-intl/blob/c6e89c5290c2ce2352926548d16df0bc7b4585e3/packages/use-intl/src/react/useFormatter.tsx).

apps/researcher/src/lib/date-formatter/hooks.tsx
barbarah: It can be `.ts`. However, I have named most, but not all, files with hooks `.tsx`.

I chose `.tsx` because hooks are only conceptually valid in a react component and only technically allowed in react components.

However, there is no jsx in this file, so it's not needed.

Researching this in our codebase, I see we are not consistent in file extensions for hooks. I will rename this one to `.ts`, but want to think/read some more before renaming all hooks files.

apps/researcher/src/lib/date-formatter/create-formatter.test.ts
barbarah: This is correct. Dates before 0 use the translations, and dates after 0 use the date format. So the two suggested tests will never happen. I can add a `expect(t).not.toHaveBeenCalled();` and `expect(formatDateTime).not.toHaveBeenCalled();` to the extisting tests to make that more clear.

sdevalk: Yes, I think that would be a good thing, for clarity. But if you think that the tests look silly and don't add value, you could instead add a comment explaining why the tests do not cover certain (logical) use cases.

sdevalk (approved):

PR #461, opened 2024-02-07
Community avatar square constrains

Doppen:
The avatar of the community had the shape of of the image but should be square.


barbarah (approved):
Nice, looks a lot better

Doppen:
Tnx

PR #468, opened 2024-02-08
Quick fix: language changes will result in hidden buttons

barbarah:


barbarah:
@sdevalk I think this is a quick fix for the disappearing buttons after the language change. For me, this works in Firefox and Chrome, but my environments behave slightly differently than on your machine/account. So I would like you to test is too.

I have also created a support ticket so the Clerk team may have different solutions. https://discord.com/channels/856971667393609759/1205175724726812682

sdevalk (approved):
👍🏻

PR #481, opened 2024-02-15
layout loginpage

Doppen:


Doppen:
Only changed the layout. Simply in the middle of the page.

barbarah (approved):

PR #488, opened 2024-02-21
feat: refactor api

sdevalk:
This PR refactors the code in the `api` folders of the `dataset-browser` and `researcher`: all code in these folders has been moved to a new, dedicated package, `api`. This makes it easier to manage the code and makes the code of the apps a bit cleaner, focussed on the frontend instead of also on the API.

Some notes:
1. The PR is huuuuuuuuge, but it's mostly just a move of code from two folders into one.
1. The API code about 'persons' has been renamed to 'constituents'.
1. ⚠️ All changes are/should be backward-compatible, with one known exception: the birth and death dates of constituents are no longer a single date but a range (a start date and an end date). This is due to a change that Kinsuk made in the backing datamodel. This backward-incompatibility means that the PR currently fails and that the frontend code of constituents needs to be updated first before this PR can be merged.
2. The sort by 'relevance' was still a part of the APIs of constituents and datasets, but was removed from objects some weeks ago. For consistency this PR also removes the sort by relevance for constituents and datasets.
3. The new `api` package still contains some redundant code (i.e. it's not DRY). In a future PR I'll remove this redundancy


barbarah (approved):
I have tested everything. I all still works. I have added a commit to remove the persons pages.

PR #489, opened 2024-02-21
Fallback image in gallery and object card components

barbarah:
Ticket: https://github.com/colonial-heritage/sprints/issues/391

Design: https://gui-prototype.colonialcollections.nl/v2/search.html

This component can also easily be updated to have a loading state.

### DIFF
--- a/apps/researcher/src/components/image-with-fallback.tsx
+++ b/apps/researcher/src/components/image-with-fallback.tsx
@@ -1,0 +1,27 @@
+'use client';
+
+import React, {useState} from 'react';
+import Image, {ImageProps} from 'next/image';
+import {useTranslations} from 'next-intl';
+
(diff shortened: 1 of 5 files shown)


apps/researcher/src/components/image-with-fallback.tsx
```diff
@@ -0,0 +1,27 @@
...
+      <div className="border-b border-neutral-200 p-4">{t('errorText')}</div>
+    );
+  }
+
+  return (
+    <Image
+      {...rest}
+      alt={alt}
+      src={src}
+      onError={() => {
```
sdevalk, line 22: Would `{() => setHasError(true)}` also work, the condensed notation?

sdevalk (approved):

PR #490, opened 2024-02-22
Add a map to the object details

barbarah:
Ticket: https://github.com/colonial-heritage/sprints/issues/378

Design: https://gui-prototype.colonialcollections.nl/v2/object.html#provider

I also moved the info block to the bottom of the page.


apps/researcher/src/app/[locale]/objects/[id]/map/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+  try {
+    const addressString = [
+      address.streetAddress,
+      address.postalCode,
+      address.addressLocality,
+      address.addressCountry,
+    ]
+      .filter(n => n)
+      .join(',');
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/objects/[id]/map/index.tsx:27]: You could place these lines outside the `try...`, to make clear that it doesn't need to be handled by the `catch`.

sdevalk (approved):
Nice! I've added one suggestion.

PR #496, opened 2024-02-27
Refactor researcher app to use ECMAScript modules (ESM)

barbarah:
This pull request details the transition of the research app and dataset browser from CommonJS to ECMAScript Modules (ESM). The motivation behind this move is to align our projects with modern JavaScript standards, improve module interoperability, and prepare our codebase for future scalability and performance enhancements.

Next.js 13+ supports ESM simply by adding type: `module` to your package.json and renaming next.config.js → next.config.mjs.

# Step 1: Type `module` in `package.json`

Node.js allows JavaScript to be interpreted as an ES module via the package.json ["type"](https://nodejs.org/api/packages.html#type) field with a value of "module". Most config files are in CommonJS, so they need to be updated. If possible, I used `.ts` for config files. For others, alternative formats were chosen, as detailed below.

- **Next.js**: Now uses `next.config.mjs`, as `.ts` is not supported yet ([Next.js Pull Request #57656](https://github.com/vercel/next.js/pull/57656)).
- **ESLint**: Transitioned to `.eslintrc.json` from `.eslintrc.js` based on [ESLint's documentation](https://eslint.org/docs/latest/use/configure/configuration-files).
- **PostCSS**: Changed to `postcss.config.json` from `postcss.config.js`, in line with Next.js [PostCSS configuration guidelines](https://nextjs.org/docs/pages/building-your-application/configuring/post-css).
- **Tailwind**: Updated to `tailwind.config.ts`.
- **Jest**: Due to an issue with ESM ([Jest Issue #11453](https://github.com/jestjs/jest/issues/11453)), moved to `.cjs` format for `jest.setup.js` and `jest.config.js`.

# Step 2: Updating the typescript config

For Next.js, the recommended settings are `module: esnext` and `moduleResolution: bundler`:

`module: esnext` instructs TypeScript to output ESM syntax, similar to nodenext, but it's a more general setting that isn't tailored exclusively to Node.js. It's suitable for code running in various environments, including browsers and bundlers (like Webpack, which Next.js uses internally).

`moduleResolution: bundler` tells TypeScript that module resolution will be handled by a module bundler, which can apply additional transformations or optimizations to the modules beyond what Node.js's native ESM resolution does.

These changes needed some minor updates:

1. `src/i18n.ts` gave errors; I updated it based on the newest example code of `next-intl`
2.  package `iso-639-1-dir` showed some typescript errors, I replaced it with the package it was forked from (and now updated) `iso-639-1`.


apps/dataset-browser/src/i18n.ts
```diff
@@ -1,5 +1,17 @@
+import {notFound} from 'next/navigation';
 import {getRequestConfig} from 'next-intl/server';
+import {locales} from './navigation';
+import {LocaleEnum} from '@/definitions';
 
-export default getRequestConfig(async ({locale}) => ({
-  messages: (await import(`./messages/${locale}/messages.json`)).default,
-}));
+export default getRequestConfig(async ({locale}) => {
+  if (!locales.includes(locale as LocaleEnum)) notFound();
```
sdevalk, line 7: For consistency, would it be possible to use braces?:

```
if () {

}
```

apps/dataset-browser/tailwind.config.ts
```diff
@@ -0,0 +1,7 @@
+import sharedConfig from 'tailwind-config/tailwind.config';
+
+const config = {
```
sdevalk, line 3: Would it be possible to add `export default` to this line, so as to make immediately clear that it's being exported as a default?

packages/tailwind-config/tailwind.config.ts
```diff
@@ -1,12 +1,13 @@
-/** @type {import('tailwindcss').Config} */
-module.exports = {
+import type {Config} from 'tailwindcss';
+
+const config = {
```
sdevalk, line 3: Would it be possible to add `export default` to this line, so as to make immediately clear that it's being exported as a default?

packages/tsconfig/nextjs.json
```diff
@@ -13,6 +13,8 @@
     "noEmit": true,
     "incremental": true,
     "module": "esnext",
+    "moduleResolution": "Bundler",
```
sdevalk, line 16: Very minor, for consistency with other settings (e.g. `esnext`): use `bundler` (lowercase)?

apps/researcher/tailwind.config.ts
```diff
@@ -0,0 +1,7 @@
+import sharedConfig from 'tailwind-config/tailwind.config';
+
+const config = {
```
sdevalk, line 3: Would it be possible to add `export default` to this line, so as to make immediately clear that it's being exported as a default?

packages/ui/tailwind.config.js
sdevalk, line 1: Is it possible to use the ESM syntax in this file (e.g. with a `.ts` extension and `import` instead of `require`)?

apps/dataset-browser/tailwind.config.ts
```diff
@@ -0,0 +1,7 @@
+import sharedConfig from 'tailwind-config/tailwind.config';
```
sdevalk, line 1: 1. The name `tailwind-config/tailwind.config` sounds a bit redundant. Is is possible to use e.g. just `from 'tailwind-config'`?
2. The names `sharedConfig` and `config` sound a bit confusing to me; they're quite similar. Would the following work (untested)?

```
import config from 'tailwind-config/tailwind.config';

export default {
  presets: [config],
};
```

sdevalk (commented):
Great work!

I've added some questions and comments.

packages/tailwind-config/tailwind.config.ts
barbarah: Only if we change this eslint rule: https://github.com/import-js/eslint-plugin-import/blob/v2.28.1/docs/rules/no-anonymous-default-export.md

packages/tailwind-config/index.ts
```diff
@@ -87,4 +88,4 @@ module.exports = {
     require('@tailwindcss/typography'),
     require('@tailwindcss/aspect-ratio'),
```
sdevalk, line 89: Sorry, I had overlooked this one during the previous review. Is it possible to use `import` here instead of `require`?

sdevalk (commented):
Nice work! I've added two questions.

packages/tailwind-config/index.ts
barbarah: For this pull request, I focused only on the apps, but now I have them working. I can also go through the files in the packages.

barbarah:
Renaming the config packages did give a problem. Cypress gave a weird error after renaming the package `tsconfig` to `@colonial-collections/ts-config`. This disappeared after using a relative path for the typescript `extends`. I think this has to do with the `ts-node` Cypress uses; there are multiple `ts-node` issues with `extends`. Looking at the issues, I believe `ts-node` works better with a relative path. I see we already use relative paths for our base config: https://github.com/colonial-heritage/colonial-collections/blob/main/packages/tsconfig/base.json.

PR #508, opened 2024-03-08
feat: constituents searcher for adding provenance information

sdevalk:
This PR is a first step to allow users to add their own provenance records to the records of the data providers (see the design: https://gui-prototype.colonialcollections.nl/object.html, button "Add provenance", fields "Transferred from" and "Transferred to").

This PR adds two new classes:

1. `packages/api/src/enrichments/searcher-constituents-wikidata.ts`. This class allows us to search for the constituents in Wikidata. Beware: Wikidata is slow - a search can take several seconds.
1. `packages/api/src/enrichments/searcher-constituents-datahub.ts`. This class allows us to search for the constituents in the database of the Datahub.

The frontend should get some logic to determine when to search in these sources. The order, according to the requirements of Gertjan, is:

1. First search for constituents in Wikidata
2. If no results are found, search for constituents in the Datahub.

We'll probably need to revisit this as soon as the first frontend implementation is in place. For example: what if Wikidata returns just one or two results, will that suffice? Or should we then also search in the Datahub and add its results to those of Wikidata?

In its slipstream this PR also updates some tests and refactors the way that we query Elasticsearch a little bit, to be more DRY.


packages/api/src/enrichments/searcher-constituents-wikidata.integration.test.ts
```diff
@@ -0,0 +1,68 @@
...
+describe('search', () => {
+  it('finds constituents that match the query in the specified locale', async () => {
+    const result = await constituentSearcher.search({
+      query: 'rembrandt',
+      locale: 'nl',
+      limit: 3,
+    });
+
+    expect(result).toStrictEqual({
+      things: [
```
barbarah, line 21: I didn't expect the word `things` here. I understand it comes from the type `Thing`. Most APIs use the term `data` for the data response, but that is even more generic than a `thing`. `constituents` is also an option, but maybe that is not generic enough. So I'm not sure about the best option. you can ignore this comment.

barbarah (approved):
I tested the code in the frontend, everything looks good. I added one comment, that can be ignored if you want.

packages/api/src/enrichments/searcher-constituents-wikidata.integration.test.ts
sdevalk: Thanks! Yes, `things` could sound a bit strange. It's a term from the base data model that the API uses, https://schema.org/Thing. The term `constituents` is an option, but I looked for a generic name that also allows other types of 'things', such as locations.

PR #509, opened 2024-03-11
feat: add locale support for types and materials

sdevalk:
This PR adds locale support for the fields 'Types' and 'Materials'. This allows users to see the values of these fields in either Dutch or English.


barbarah (approved):
Nice feature!

PR #516, opened 2024-03-15
feat: search Geonames for locations

sdevalk:
This PR adds support for searching for locations in GeoNames. This functionality can be used in the frontend, in the form for adding new provenance information (field "Location"): https://gui-prototype.colonialcollections.nl/object.html

The GeoNames API requires a `username` for making authenticated requests. This value can be found in [Vercel](https://vercel.com/colonial-heritage/colonial-collections-researcher/settings/environment-variables). [There are usage limits in place, though](https://www.geonames.org/export/):

> 10'000 [credits](https://www.geonames.org/export/credits.html) daily limit per application (identified by the parameter 'username'), the hourly limit is 1000 [credit](https://www.geonames.org/export/credits.html)s. A credit is a web service request hit for most services. An [exception](https://www.geonames.org/export/webservice-exception.html) is thrown when the limit is exceeded.

The current `username` is a free account for testing. We can upgrade to a paid account if we run out of credits.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,7 @@ Create the file `apps/researcher/.env.production.local` and set the endpoint URL
     SENDGRID_API_KEY=
     TO_EMAIL_ADDRESS=
     FROM_EMAIL_ADDRESS=
+    GEONAMES_USERNAME=
 
 Then run:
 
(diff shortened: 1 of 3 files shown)


barbarah (approved):
It looks good. It's nice that all the searchers work the same way.

PR #525, opened 2024-03-20
Update "add object to list" button

barbarah:
This pull request contains three changes to "add object to list" button.

1. Show a button with a text if the user is not logged in OR has no communities. The second part of the ticket: https://github.com/colonial-heritage/sprints/issues/416
2. https://github.com/colonial-heritage/sprints/issues/417
3. Show a loading text instead of "no lists" text when the lists are loading, no ticket, just something I really wanted to change


apps/researcher/cypress/e2e/object.cy.ts
```diff
@@ -59,6 +59,18 @@ describe('Object details page', () => {
         cy.getBySel('selectedFilter').should('have.length', 2);
       });
   });
+
+  it('shows an text when hovering the add-to-list-button', () => {
```
sdevalk, line 63: Minor: 'shows a text...'

sdevalk (approved):

PR #527, opened 2024-03-21
Add ReadMoreText component to display long descriptions

barbarah:
Ticket: https://github.com/colonial-heritage/sprints/issues/369

### DIFF
--- a/apps/researcher/src/components/read-more-text.tsx
+++ b/apps/researcher/src/components/read-more-text.tsx
@@ -1,0 +1,32 @@
+'use client';
+
+import {useState} from 'react';
+
+interface ReadMoreTextProps {
+  text?: string;
+  maxLength?: number;
+}
+
+export function ReadMoreText({text, maxLength = 300}: ReadMoreTextProps) {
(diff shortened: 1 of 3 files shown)


apps/researcher/src/components/read-more-text.tsx
```diff
@@ -0,0 +1,32 @@
...
+    return null;
+  }
+
+  if (text.length <= maxLength) {
+    return <p>{text}</p>;
+  }
+
+  return (
+    <p>
+      {isExpanded ? text : `${text.slice(0, maxLength)}... `}
```
sdevalk, line 23: Instead of slicing, would it be possible to hide and show the full text with Tailwind's [truncate](https://tailwindcss.com/docs/text-overflow#truncate)?

sdevalk (approved):

apps/researcher/src/components/read-more-text.tsx
barbarah: We used tailwinds `line-clamp` before; that works better for multi-line purposes. The problem is showing the "read-more" button conditionally. There is not a "if text is clamp" class.

PR #565, opened 2024-04-12
feat: create persistent IRI

sdevalk:
This PR adds a little helper function for creating persistent IRIs. We can use this helper e.g. when creating communities and adding the IRIs to the metadata of the communities in Clerk.


barbarah (approved):
This looks good, I will create the next step. Adding the url to the public metedata on create

PR #567, opened 2024-04-16
feat: store and retrieve community of creator from nanopublication

sdevalk:
This PR expands the `enricher` package: in addition to storing and retrieving the `creator` of an enrichment, it's now possible to also store and retrieve the community to which the creator belongs [1].

Some notes:
1. The data model used for storing the community information in the backing nanopublication could change - Kinsuk is still looking into this. This shouldn't impact the frontend, though
2. The PR also harmonizes some names: the `enricher` package referred to 'actor', 'agent' and 'constituent' - different names for essentially the same thing. It now uses 'actor' in all places
3. The PR also improves some tests
4. The code of the package could become more DRY. I'll do this in a future PR, if the functionality of this PR is stable

[1] For storing the community, see e.g. https://github.com/colonial-heritage/colonial-collections/blob/feat-enricher-community/packages/enricher/src/creator.integration.test.ts#L32-L34. For retrieving the community, see e.g. https://github.com/colonial-heritage/colonial-collections/blob/feat-enricher-community/packages/enricher/src/objects/fetcher.integration.test.ts#L96-L98

### DIFF
--- a/apps/researcher/src/app/[locale]/objects/[id]/metadata.tsx
+++ b/apps/researcher/src/app/[locale]/objects/[id]/metadata.tsx
@@ -13,7 +13,7 @@ import {
   XMarkIcon,
 } from '@heroicons/react/24/outline';
 import type {
-  Agent,
+  Actor,
   HeritageObjectEnrichmentType,
 } from '@colonial-collections/enricher';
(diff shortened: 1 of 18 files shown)


barbarah (approved):
Nice! I can incorporate this into the form i'm making (adding provenance).

PR #568, opened 2024-04-17
Add Provenance event creation form with new reusable components

barbarah:
Ticker: https://github.com/colonial-heritage/sprints/issues/345
Design: https://gui-prototype.colonialcollections.nl/object.html

This pull request introduces a new form for creating provenance events. This form is significant as it incorporates multiple new reusable form components designed to adhere to the DRY principle.

### Key Features:
1. **SearchSelector:** This is a new component that allows users to search through multiple APIs and select results. This is our first implementation of client-side fetching using `SWR`. I've added the necessary API routes to support this functionality.
2. **EDTF Input:** This component accepts a year or an exact date. The design could be better. Bas is going to update it.
3. **Select:** A straightforward selector component that supports descriptions for options.

### Implementation Notes:
- The new components are currently exclusive to this form. There's a plan to extend these components to other forms in the application to maintain consistency and reusability. I did update the enrichment form slightly so both forms can use the same language selector. But a full refactor so the form uses all new components is still needed.
- Given our limited use of client-side data fetching, opting for `SWR` over more complex solutions like `react-query` aligns with our need for simplicity and efficiency.
- **Sorting Update:** The provenance events in the data table are now sorted by actual date rather than by `startsAfter` and `endsBefore`. This change was necessary because the newly added events do not have `startsAfter` and `endsBefore` fields. This approach simplifies the sorting logic.

I did not add the community selector yet, this will be the next step.


apps/researcher/src/app/[locale]/api/datahub/route.ts
```diff
@@ -0,0 +1,19 @@
...
+const datahubConstituentSearcher = new DatahubConstituentSearcher({
+  endpointUrl: env.SEARCH_ENDPOINT_URL as string,
+});
+
+export async function GET(request: NextRequest) {
+  const searchParams = request.nextUrl.searchParams;
+  const query = searchParams.get('query');
+  const result = await datahubConstituentSearcher.search({
+    query: query || '',
+    locale: 'en',
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/api/datahub/route.ts:14]: Should this be an option/a parameter?

apps/researcher/src/app/[locale]/api/geonames/route.ts
```diff
@@ -0,0 +1,20 @@
...
+  endpointUrl: 'http://api.geonames.org',
+  username: env.GEONAMES_USERNAME as string,
+});
+
+export async function GET(request: NextRequest) {
+  const searchParams = request.nextUrl.searchParams;
+  const query = searchParams.get('query');
+  const result = await geoNamesLocationSearcher.search({
+    query: query || '',
+    locale: 'en',
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/api/geonames/route.ts:15]: Should this be an option/a parameter?

apps/researcher/src/app/[locale]/api/wikidata/route.ts
```diff
@@ -0,0 +1,19 @@
...
+  endpointUrl: 'https://query.wikidata.org/sparql',
+});
+
+export async function GET(request: NextRequest) {
+  const searchParams = request.nextUrl.searchParams;
+  const query = searchParams.get('query');
+
+  const result = await datahubConstituentSearcher.search({
+    query: query || '',
+    locale: 'en',
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/api/wikidata/route.ts:14]: Should this be an option/a parameter?

apps/researcher/src/app/[locale]/objects/[id]/(provenance)/add-form.tsx
```diff
@@ -0,0 +1,455 @@
...
+          z.string().refine(isEdtf, {message: t('invalidStartDate')}),
+        ]),
+        endDate: z.union([
+          z.literal(''),
+          z.string().refine(isEdtf, {message: t('invalidEndDate')}),
+        ]),
+      })
+      .refine(
+        ({startDate, endDate}) => {
+          if (!isEdtf(startDate) || !isEdtf(endDate)) return true;
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/objects/[id]/(provenance)/add-form.tsx:108]: To be sure: is the OR condition correct (it's not an AND)?

apps/researcher/src/app/[locale]/objects/[id]/(provenance)/sort-events.ts
```diff
@@ -1,73 +1,19 @@
...
-  // Step 1: Initialize the `graph` and `incomingEdgeCount` map
-  // Input: events array.
-  // Operation: For each event, create an empty array in the graph,
-  // and set the incoming edge count to 0.
-  // Output: `graph` and `incomingEdgeCount` maps initialized with each event id.
-  events.forEach(event => {
-    graph.set(event.id, []);
-    incomingEdgeCount.set(event.id, 0);
-  });
+    // If endDate is missing, consider it as the smallest possible value
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/objects/[id]/(provenance)/sort-events.ts:12]: Or is it the highest possible value? If an end date is missing then, semantically, the event is still ongoing.

apps/researcher/src/app/[locale]/objects/[id]/user-enrichment-form.tsx
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/objects/[id]/user-enrichment-form.tsx:1]: There are some GitHub warnings in this file.

apps/researcher/src/components/form/edtf/converter.test.ts
```diff
@@ -0,0 +1,83 @@
...
+    expect(result).toEqual('2022');
+  });
+
+  it('returns the correct date string when only year is provided', () => {
+    const yyyy = '2022';
+    const result = toDateString(yyyy);
+    expect(result).toEqual('2022');
+  });
+
+  it('returns an negative year if year is negative', () => {
```
sdevalk, line 38: Minor: `returns a negative`

apps/researcher/src/components/form/edtf/converter.test.ts
```diff
@@ -0,0 +1,83 @@
...
+    const yyyy = '-500';
+    const result = toDateString(yyyy);
+    expect(result).toEqual('-0500');
+  });
+
+  it('returns the correct date string when year is negative and other components are provided', () => {
```
sdevalk, line 44: Perhaps be explicit about 'other components'? (what are these, and what are 'components'?)

apps/researcher/src/components/form/edtf/converter.test.ts
```diff
@@ -0,0 +1,83 @@
...
+});
+
+describe('fromDateString', () => {
+  it('returns the correct date components when a valid date string is provided', () => {
+    const dateString = '2022-10-15';
+    const result = fromDateString(dateString);
+    expect(result).toEqual({yyyy: '2022', mm: '10', dd: '15'});
+  });
+
+  it('returns an empty object when an empty string is provided', () => {
```
sdevalk, line 60: Looking at the `result` I'd say that the object isn't empty (`{}`) but that the values of the properties are empty?

apps/researcher/src/components/form/edtf/converter.test.ts
```diff
@@ -0,0 +1,83 @@
...
+    expect(result).toEqual({yyyy: '-500', mm: '10', dd: '15'});
+  });
+
+  it('returns the correct date components when leading zeros are present', () => {
+    const dateString = '0001-01-01';
+    const result = fromDateString(dateString);
+    expect(result).toEqual({yyyy: '1', mm: '1', dd: '1'});
+  });
+
+  it('returns the correct date components when only year is provided', () => {
```
sdevalk, line 78: Minor: `when only a year`?

apps/researcher/src/components/form/edtf/converter.ts
```diff
@@ -0,0 +1,36 @@
+export function toDateString(yyyy = '', mm = '', dd = ''): string {
```
sdevalk, line 1: Is the `string` as return type required (i.e. TypeScript cannot infer this)?

apps/researcher/src/components/form/edtf/converter.ts
```diff
@@ -0,0 +1,36 @@
...
+  const mmPadded = mm.padStart(2, '0');
+  const ddPadded = dd.padStart(2, '0');
+
+  if (ddPadded === '00' && mmPadded === '00') return yyyyPadded;
+  if (ddPadded === '00') return `${yyyyPadded}-${mmPadded}`;
+  if (mmPadded === '00') return yyyyPadded;
+  return `${yyyyPadded}-${mmPadded}-${ddPadded}`;
+}
+
+export function fromDateString(dateString: string): {
```
sdevalk, line 19: Perhaps omit the trailing `:`?

apps/researcher/src/components/form/edtf/converter.ts
```diff
@@ -0,0 +1,36 @@
...
+  yyyy: string;
+  mm: string;
+  dd: string;
+} {
+  const parts = dateString.split(/(?<=^-?\d{1,4}|\d{2})-/);
```
sdevalk, line 24: In the unit tests these `parts` seem to be called `components`. Use one term, for consistency?

sdevalk (commented):
Nice work! I've added some questions.

This is a huge PR, so I might have missed some details. I can take a closer look as soon as the PR is deployed.

sdevalk:
In response to:

> Sorting Update: The provenance events in the data table are now sorted by actual date rather than by startsAfter and endsBefore. This change was necessary because the newly added events do not have startsAfter and endsBefore fields. This approach simplifies the sorting logic.

Yes, this makes sense. It could be, though, that we'll eventually need a (complicated...) sorting combination of both: `startsAfter` + `endsBefore` and the actual dates, depending on how data providers provide their data.

apps/researcher/src/app/[locale]/objects/[id]/(provenance)/add-form.tsx
Comment (barbarah) [reply apps/researcher/src/app/[locale]/objects/[id]/(provenance)/add-form.tsx:108]: Yes, if one is not a valid date, then there is no range. So a range validation is not needed.

A not-valid edtf can mean that it is empty (which is valid) or has a weird value, in which case the validation will be done on the start/end date level.

apps/researcher/src/app/[locale]/objects/[id]/(provenance)/sort-events.ts
Comment (barbarah) [reply apps/researcher/src/app/[locale]/objects/[id]/(provenance)/sort-events.ts:12]: This is true; I will change it.

apps/researcher/src/components/form/edtf/converter.ts
barbarah: Good point, I will go for `parts`

apps/researcher/src/components/form/edtf/converter.test.ts
```diff
@@ -0,0 +1,83 @@
+import {expect} from '@jest/globals';
+import {fromDateString, toDateString} from './converter';
+
+describe('toDateString', () => {
+  it('returns an empty string when no date components are provided', () => {
```
sdevalk, line 5: For consistency, also use the term 'parts' here (rather than 'components')? This also applies to the test underneath.

sdevalk (approved):
Perfect! Just one comment :)

PR #569, opened 2024-04-17
Add iri to community metadata on creation

barbarah:
In this pull request, the iri is added to the community after creation.

We create a community using the `openCreateCommunity` modal provided by Clerk. The `openCreateCommunity` has a `afterCreateOrganizationUrl` prop.

This is set to `/communities/${organization.slug}/created`.

The URI is set on this page, and the user is redirected to the community page.

I wouldn't say I like logic based on navigation to a URL, but this is the only entry `openCreateCommunity` gives us. The alternative is not using the clerk modal but creating the modal ourselves and creating the community using the Clerk API; this gives us more options, but it is more work.

### DIFF
--- a/apps/researcher/src/lib/community/actions.ts
+++ b/apps/researcher/src/lib/community/actions.ts
@@ -172,7 +172,7 @@ interface UpdateCommunityIriProps {
   iri: string;
 }
 
-export async function addUriToCommunity({id, iri}: UpdateCommunityIriProps) {
+export async function addIriToCommunity({id, iri}: UpdateCommunityIriProps) {
   noStore();
 
   const community = await getCommunityById(id);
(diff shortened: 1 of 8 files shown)


apps/researcher/src/lib/community/actions.ts
```diff
@@ -154,12 +154,38 @@ export async function updateCommunity({
...
 
   return organizationToCommunity(organization);
 }
+
+interface UpdateCommunityIriProps {
+  id: string;
+  iri: string;
+}
+
+export async function addUriToCommunity({id, iri}: UpdateCommunityIriProps) {
```
sdevalk, line 175: Minor: for consistency use `addIriToCommunity()` (Iri instead of Uri)?

sdevalk (approved):
Nice! I've added one suggestion.

PR #577, opened 2024-04-22
Fix warning LocalizedMarkdown

barbarah:
The `LocalizedMarkdown` component was created for server components. But it is also now used in client components, which warns that "async/await is not yet supported in Client Component.". I added a temporary fix by wrapping the `LocalizedMarkdown` within client components with a `<Suspense>`. With the ` <Suspense>` the component `LocalizedMarkdown` loads, but the warning was still visible.

I fix this by removing the 'async' in `LocalizedMarkdown`.

### DIFF
--- a/apps/researcher/src/app/[locale]/objects/[id]/(provenance)/add-form.tsx
+++ b/apps/researcher/src/app/[locale]/objects/[id]/(provenance)/add-form.tsx
@@ -2,7 +2,7 @@
 
 import {Tab} from '@headlessui/react';
 import {useLocale, useTranslations} from 'next-intl';
-import {Fragment, Suspense, useMemo, useState} from 'react';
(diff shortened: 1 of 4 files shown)


apps/researcher/src/app/[locale]/objects/[id]/(provenance)/add-form.tsx
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/objects/[id]/(provenance)/add-form.tsx:1]: There is a GitHub Actions warning in this file

apps/researcher/src/app/[locale]/objects/[id]/user-enrichment-form.tsx
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/objects/[id]/user-enrichment-form.tsx:1]: There is a GitHub Actions warning in this file

sdevalk (approved):
Nice! I've added two comments.

PR #585, opened 2024-04-24
Update Provenance form and data table

barbarah:
## Add motivation to the form
Users can add motivations to multiple fields. The motivations are saved as a YAML in the `description` field.

## Translate provenance options
Translate the type and qualifier to English on save. This can only be done in the server action; the frontend components cannot get translations in a different locale than the selected language.

Translate it back to the selected language in the data table. If the ID is unknown, use the saved name. This could happen if nanopubs are saved outside our application.

## Show all props in the data table
With the new fields, the data table had a lot of logic, so we extracted the logic and moved it to a `transformEvents` function.


apps/researcher/src/app/[locale]/objects/[id]/(provenance)/actions.ts
```diff
@@ -56,7 +67,27 @@ export async function addProvenanceEnrichment({
...
+    locale: 'en',
+    namespace: 'QualifierSelector',
+  });
+
+  const tEnType = await getTranslations({
+    locale: 'en',
+    namespace: 'ProvenanceEventType',
+  });
+  const motivationsWithValues = Object.entries(motivations).reduce(
+    (list, [key, motivation]) => {
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/objects/[id]/(provenance)/actions.ts:82]: Minor: the name `list` is very generic - it could be any list. To make its meaning clearer, use a functional name instead?

apps/researcher/src/messages/en/messages.json
```diff
@@ -227,6 +228,13 @@
     "possibly": "Possibly",
     "probably": "Probably"
   },
+  "MotivationInput": {
+    "openButton": "Add motivation",
+    "editButton": "Edit motivation",
+    "description": "In this motivation field you can add an extra motivation about and only about the above field. It will only be display at the provenance section and will affect the search results.",
```
sdevalk, line 234: 1. `displayed`?
2. What does 'and will affect the search results.' mean?

barbarah: Good question, I copied this from the design. I will ask Bas

barbarah: For now I have completely rewritten this description. Without saying anything about what we do with this information. This can be updated by Camiel if needed.

sdevalk (approved):

PR #587, opened 2024-04-25
Upgrading to Clerk Core 2

barbarah:
Part of: https://github.com/colonial-heritage/colonial-collections/pull/587

In April 2024, Clerk updated their SDKs to include Core 2. This new core includes:

- 💅 Redesigned and improved component UIs.
- 🔪 The 401 & "flash of white page" that was visible sometimes when syncing auth state was eliminated.
- 🧹 A large number of smaller housekeeping changes, bugfixes, and improvements.
- 📊 [Telemetry](https://clerk.com/docs/telemetry) to allow us to better determine feature usage and priority (easy opt-out if you don't like this)

## Implementation

- I followed the upgrade guild
- There where a also lot changes in the translations files, I updated them with two commits, the first was a full replace with there translations. The second was an update with our changes. So if you want to check what I changed see commit: https://github.com/colonial-heritage/colonial-collections/commit/f1e4c4c4bab30c980bb637a52101674fa3f6b73c

### DIFF
--- a/apps/researcher/cypress/global.d.ts
+++ b/apps/researcher/cypress/global.d.ts
@@ -23,7 +23,7 @@ declare namespace Cypress {
 
   interface ApplicationWindow {
     Clerk: {
-      isReady: () => boolean;
+      loaded: boolean;
       client: {
         signIn: {
           create: (params: {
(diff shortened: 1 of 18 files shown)


sdevalk (approved):
👍🏻

barbarah:
We decided not to update Clerk for now.

PR #590, opened 2024-04-26
fix: broaden search results

sdevalk:
This PR changes the way that we search in GeoNames: rather than limiting the results to specific feature codes (e.g. places, countries, administrative divisions), we now search for all types of geographical entities. This could result in a low precision, though. If that's the case we can revisit the use of feature codes.


barbarah (approved):
This looks much better! If I search for 'Amsterdam' the first result is in The Netherlands. If I search for 'Batavia' I also see 'Jakarta, Indonesia'.

PR #609, opened 2024-05-14
Use user iri for saving enrichments and provenance events

barbarah:
This pull request uses the implementation of https://github.com/colonial-heritage/colonial-collections/pull/608

- Remove all code for custom user iri (attributionId)
- Create a custom `useUser` hook instead of the hook from '@clerk/nextjs'. Add iri to the user object.
- Update all code using `useUser` from Clerk
- Remove the custom hook `useUserCommunities`, use the new `useUser`


sdevalk (approved):
Nice!

PR #628, opened 2024-05-29
Let the user chose an imageFetchMode before searching

barbarah:
Based on design: https://gui-prototype.colonialcollections.nl/search-empty.html

### DIFF
--- a/apps/researcher/src/app/[locale]/objects/initial-image-fetch-mode.tsx
+++ b/apps/researcher/src/app/[locale]/objects/initial-image-fetch-mode.tsx
@@ -5,7 +5,7 @@ import {Field, Label, Radio, RadioGroup} from '@headlessui/react';
 import {useTranslations} from 'next-intl';
 
 const options = [
-  {translationKey: 'hideImages', value: ImageFetchMode.None},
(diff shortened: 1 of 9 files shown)


apps/researcher/src/app/[locale]/objects/initial-image-fetch-mode.tsx
```diff
@@ -0,0 +1,51 @@
+'use client';
+
+import {ImageFetchMode, useListStore} from '@colonial-collections/list-store';
+import {Field, Label, Radio, RadioGroup} from '@headlessui/react';
+import {useTranslations} from 'next-intl';
+
+const options = [
+  {translationKey: 'hideImages', value: ImageFetchMode.None},
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/objects/initial-image-fetch-mode.tsx:8]: To be sure: is the site 'hiding' images, or 'not loading' them? The text says the former, but I would expect the latter.

sdevalk (approved):

apps/researcher/src/app/[locale]/objects/initial-image-fetch-mode.tsx
Comment (barbarah) [reply apps/researcher/src/app/[locale]/objects/initial-image-fetch-mode.tsx:8]: Good point, I will call this 'no images'

PR #632, opened 2024-05-31
simplify the layout structure

barbarah:
This is just a small refactor I wanted to do to simplify the code, and it probably also optimised the caching. 

In the past there used to be a design with different background colors that needed to be on the `body`-tag. That is why I added a `layout.tsx` for each path with a `body`-tag and a `BaseLayout` component. But the design has changed, and there is no need for colors on the `body`-tag level. That is why I now moved the `body`-tag and `BaseLayout` content to the root layout. And removed the `body`-tag and `BaseLayout` components from all other layouts. If a layout only contained a `body`-tag and `BaseLayout` component I removed the layout.


apps/researcher/src/app/layout.tsx
```diff
@@ -7,6 +7,6 @@ type Props = {
 // Even though this component is just passing its children through, the presence
 // of this file fixes an issue in Next.js 13.4 where link clicks that switch
```
sdevalk, line 8: Minor: we're using Next 14.x.x nowadays. Should we update the comment?

sdevalk (approved):

PR #645, opened 2024-06-07
Set list defaults correctly

barbarah:
The defaults for `sortBy`, `limit`, `view`, and `imageFetchMode` were not always set correctly, and there were some differences. Without changing the logic too much, I tried to set the defaults all the same way and use the same default variable for each list to prevent different values being used within the same list.

Details of the implementation

- `imageFetchMode` is not set to `ImageFetchMode.None` to prevent loading of images. Instead, in the object card, I wait for `isInitialized` before trying to load the images.
- There was no default limit; this was a problem because the communities list used a different value. The wrong limit was added to the URL search parameters when navigating back, causing unexpected behavior. To address this, I added a default for limit, which ensures a consistent and predictable user experience.
- The default sortBy was `RelevanceDesc`, but we don't use relevance for most lists, so I changed this to `NameAsc`.

### DIFF
--- a/apps/researcher/src/app/[locale]/communities/settings.ts
+++ b/apps/researcher/src/app/[locale]/communities/settings.ts
@@ -1,0 +1,1 @@
+export const limit = 12;
--- a/apps/researcher/src/app/[locale]/communities/settings.ts
+++ b/apps/researcher/src/app/[locale]/communities/settings.ts
@@ -1,1 +1,1 @@
-export const limit = 12;
+export const itemsPerPageLimit = 12;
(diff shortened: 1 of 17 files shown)


apps/researcher/src/app/[locale]/communities/settings.ts
```diff
@@ -0,0 +1 @@
+export const limit = 12;
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/communities/settings.ts:1]: The word `limit` is quite generic and its meaning may not be clear - what kind of limit is it? Would it be an idea to rename it and make its meaning explicit?

apps/researcher/src/app/[locale]/objects/heritage-object-card.tsx
```diff
@@ -9,9 +9,14 @@ import {ImageFetchMode} from '@colonial-collections/list-store';
 interface Props {
   heritageObject: HeritageObject;
   imageFetchMode: ImageFetchMode;
+  isInitialized: boolean;
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/objects/heritage-object-card.tsx:12]: For consideration: make clear what has been initialized?

sdevalk (approved):

PR #654, opened 2024-06-13
feat: store and retrieve Local Contexts Notices enrichments

sdevalk:
This PR expands the `enricher` package: it allows you to store and retrieve enrichments about Local Contexts Notices.

The backing data model for storing the enrichments (as nanopublications) could still change a bit, but it shouldn't really impact the way that you use the API.

### DIFF
--- a/packages/enricher/src/local-contexts-notices/definitions.ts
+++ b/packages/enricher/src/local-contexts-notices/definitions.ts
@@ -1,0 +1,35 @@
+import {basicEnrichmentBeingCreatedSchema, PubInfo} from '../definitions';
+import {z} from 'zod';
+
+// An enrichment can be about these Notices
+export enum LocalContextsNoticeEnrichmentType {
+  Authorization = 'https://localcontexts.org/notice/authorization/',
+  Belonging = 'https://localcontexts.org/notice/belonging/',
+  Caring = 'https://localcontexts.org/notice/caring/',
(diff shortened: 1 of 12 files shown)


packages/enricher/src/local-contexts-notices/definitions.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export type LocalContextsNoticeEnrichmentBeingCreated = z.infer<
+  typeof localContextsNoticeEnrichmentBeingCreatedSchema
+>;
+
+export type LocalContextsNoticeEnrichment = {
+  id: string;
+  type: LocalContextsNoticeEnrichmentType;
+  description?: string;
+  citation?: string;
```
barbarah, line 31: In the design, there is no `citation`. I'm not sure if this is relevant for a Local Context Notice

barbarah (approved):
I have a working (not finished) frontend with this API, works as expected. I have one question about a property that is not used in the design. You can choose to ignore this.

packages/enricher/src/local-contexts-notices/definitions.ts
sdevalk: Yes, you're right. I've decided to add the property to be consistent with the other enrichment types. But if it's not needed in the frontend, we can ignore it.

PR #673, opened 2024-06-25
Open hyperlinks in the markdown of the research guide in a new browse…

barbarah:
Open hyperlinks in the markdown of the research guide in a new browser window


GertjanFi (approved):
👍

PR #679, opened 2024-06-27
Show more information in local context notice form

barbarah:


Doppen:
The text labels are almost right:

- **_Notice text_** should be **_Why use this notice_**
- **_This notice will be displayed like this_** should be **_Notice text_**

Doppen:
And one last thing: Can you give the text "Select one of the Notices. Click on a label for more information." a div with `tabindex=0`

Doppen (approved):

PR #684, opened 2024-06-27
Lost accessibillity items

Doppen:
- tabindex
- text on close button
- misspelled aria label
- remove aria label on resultscard


barbarah (approved):

Doppen:
Thanks!!

PR #686, opened 2024-06-28
Compact page header

Doppen:
Blue part of the object page header more compact. Image more to top.

### DIFF
--- a/apps/researcher/src/app/[locale]/objects/[id]/page.tsx
+++ b/apps/researcher/src/app/[locale]/objects/[id]/page.tsx
@@ -109,7 +109,7 @@ export default async function Details({params}: Props) {
                   </div>
                 ))}
               </div>
-              <div className="text-sm my-4 xl:my-10 text-consortium-blue-100 flex gap-1">
+              <div className="text-sm my-4 xl:my-10 text-consortium-blue-100 md:flex gap-1">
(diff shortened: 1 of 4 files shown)


apps/researcher/src/app/[locale]/objects/[id]/page.tsx
```diff
@@ -217,9 +221,9 @@ export default async function Details({params}: Props) {
             </div>
             <LocalContextsNotices />
           </main>
-          <aside className="w-full md:w-1/3 self-stretch order-1 md:order-2  md:mx-0 md:bg-neutral-100 p-1">
+          <aside className="w-full lg:w-1/3 self-stretch flex flex-col justify-start order-1 lg:order-2">
             {galleryImages.length > 0 && (
-              <div className="flex flex-row md:flex-col gap-1 sticky top-4">
+              <div className="flex flex-row md:flex-col gap-1 sticky top-8 lg:-mt-72 z-30 md:mx-0 p-4 rounded bg-consortium-blue-400/10">
```
Comment (barbarah) [inline apps/researcher/src/app/[locale]/objects/[id]/page.tsx:226]: This index is higher then the slide out. The slide out has an index of 10, but should slide out above this image, so change of of the two.

image

barbarah (approved):
I have one comment, if this is fixed you can merge it.

PR #689, opened 2024-06-28
Refactor: Remove dot notations components

barbarah:
Initially, we decided on dot notation compound components for the generic components, but unfortunately, they do not work for server components. This pull request removes the dot notation. Headless UI also depreciated all dot notation components, so I replaced them with normal ones.

I also removed some header components we do not use anymore.


GertjanFi (approved):
👍

PR #730, opened 2024-12-02
Research guide frontend

barbarah:
This pull request updated the top level of the research guide based on the new API and design.

Top level page: 

- The identifier property is removed, so there is some logic added to determine the level
 - Sort by name
 - Use the names of the new level 1 guides
 - Separate design for the first level 1 guide
 - Show 4 level 2 guides next to each other; do not break within the level 2 guide
 - Only show level 3 guides below the level 2 guides; filter out level 1 and 2

Details page:

- Open links to Resources in new tab
- Remove the 'Description' and 'Resources' buttons
- Display the 'name variants' field/information
- Display the 'active period' field/information

### DIFF
--- a/apps/researcher/src/app/[locale]/research-guide/filterGuides.test.ts
+++ b/apps/researcher/src/app/[locale]/research-guide/filterGuides.test.ts
@@ -32,15 +32,10 @@ describe('filterLevel3Guides', () => {
 
     const filteredTopLevel = filterLevel3Guides(topLevel);
 
-    expect(filteredTopLevel.seeAlso?.[0].seeAlso?.[0].seeAlso).toEqual([
-      {id: 'level3-1'},
-      {id: 'level3-2'},
-      {id: 'level3-3'},
-    ]);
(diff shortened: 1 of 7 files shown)


barbarah:
Based on the last meeting, I have updated the design:

- All guides are sorted alphabetically.
- Changed layout top level:
     - The text and first guide box are now next to each other
     - Small alignment changes
 - Period of activity:
    - Only show years
    - Change the text if dates are not present
- Add label to the name variants
- Change the style of the links in the Markdown, they should be less visible, most are not very important.

apps/researcher/src/app/[locale]/research-guide/filterGuides.test.ts
```diff
@@ -0,0 +1,197 @@
+import {ResearchGuide} from '@colonial-collections/api/src/research-guides/definitions';
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/research-guide/filterGuides.test.ts:1]: I had forgotten to export the definitions 🥲 and just updated my PR. Would this work?

`import {ResearchGuide} from '@colonial-collections/api';`

(Also applies to the rest of this PR)

sdevalk (approved):
Nice! I've added one comment/question.

PR #733, opened 2024-12-13
Has parts frontend

barbarah:


apps/researcher/src/app/[locale]/research-guide/sortGuides.test.ts
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/research-guide/sortGuides.test.ts:1]: Minor: use dashes in the file name, instead of camelCase, for consistency with other files?

apps/researcher/src/app/[locale]/research-guide/sortGuides.ts
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/research-guide/sortGuides.ts:1]: Minor: use dashes in the file name, instead of camelCase, for consistency with other files?

apps/researcher/src/app/[locale]/research-guide/sortGuides.ts
```diff
@@ -0,0 +1,17 @@
+import {ResearchGuide} from '@colonial-collections/api';
+
+export function sortResearchGuide(topLevel: ResearchGuide): ResearchGuide {
```
Comment (sdevalk) [inline apps/researcher/src/app/[locale]/research-guide/sortGuides.ts:3]: Curious: can we omit the return type (`ResearchGuide`), i.e. is TypeScript able to infer this?

sdevalk (approved):
Nice! I've added some minor questions.

apps/researcher/src/app/[locale]/research-guide/sortGuides.ts
Comment (barbarah) [reply apps/researcher/src/app/[locale]/research-guide/sortGuides.ts:3]: Yes, we can

PR #751, opened 2025-07-04
Use a higher limit for iri check

barbarah:
The check worked, but only 7 users were updated. `getUserList` has a default limit of 10; therefore, it will not retrieve all users. I updated it to 200. There are about 140 user currently.

Docs: https://clerk.com/docs/references/backend/user/get-user-list

### DIFF
--- a/apps/researcher/src/app/[locale]/iri-check/page.tsx
+++ b/apps/researcher/src/app/[locale]/iri-check/page.tsx
@@ -6,7 +6,10 @@ export default async function IriCheckPage() {
   noStore(); // Disable caching to prevent double updates
 
   let updated = 0;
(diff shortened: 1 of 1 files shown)


jauco (approved):
can we filter out the users with an iri? Otherwise this page will stop working once 60 more users have signed up

approving it so we can apply the fix for the current users

barbarah:
Unfortunately, it is not possible to filter on metadata. By ordering with '-created_at', you always get the newest users, and we only need to check new users, which is why I felt this was good enough. But I do agree this approach is not bulletproof. Hopefully, the updated Clerk settings implementation makes this page obsolete, so for now, this is a good start.

PR #755, opened 2025-07-09
Improve Provenance segment on Object page

Doppen:
replaced some buttons to keep this segment more in line with the other segments.
https://github.com/colonial-heritage/sprints/issues/528

### DIFF
--- a/apps/researcher/src/app/[locale]/objects/[id]/(provenance)/overview.tsx
+++ b/apps/researcher/src/app/[locale]/objects/[id]/(provenance)/overview.tsx
@@ -103,7 +103,7 @@ async function AddProvenanceButton() {
 
   return (
     <SlideOutButton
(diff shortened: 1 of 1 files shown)


apps/researcher/src/app/[locale]/objects/[id]/(provenance)/overview.tsx
```diff
@@ -98,9 +103,10 @@ async function AddProvenanceButton() {
 
   return (
     <SlideOutButton
-      className="p-1 sm:py-2 sm:px-3 rounded-full text-xs bg-neutral-200/50 hover:bg-neutral-300/50 text-neutral-800 transition flex items-center gap-1"
+      className="mb-4 py-2 px-3 p-1 sm:py-2 sm:px-3 rounded-full text-xs bg-consortium-green-300 text-consortiumBlue-800 transition flex items-center gap-1 hover:bg-consortium-green-200 whitespace-pre"
```
Comment (barbarah) [inline apps/researcher/src/app/[locale]/objects/[id]/(provenance)/overview.tsx:106]: In the app, we do not use pascalCase for our classes. `consortiumBlue` does not exist, this should be `consortium-blue`

Doppen:
changed the color name

PR #757, opened 2025-07-11
Create user iri on save enrichment

barbarah:
Don't add a user IRI during sign-up; create one if an IRI is required for adding enrichments.

I tested this with a new user.

- On create, no user
- After enrichment creation, there was an IRI
- After the second enrichment creation, the IRI did not change


barbarah:
I tested this with a new user.

- On create, no user
- After enrichment creation, there was an IRI
- After the second enrichment creation, the IRI did not change

apps/researcher/src/lib/user/actions.ts
```diff
@@ -0,0 +1,24 @@
+import {clerkClient} from '@clerk/nextjs';
+import {createPersistentIri} from '@colonial-collections/iris';
+import {unstable_noStore as noStore} from 'next/cache';
+
+export async function getOrSetIri(userId: string) {
```
sdevalk, line 5: Minor: rename the function to make its meaning clearer? For example: `getIriOfUser()`?

We could also make clear in the name that the IRI is stored if it doesn't exist, but we can also regard this as an implementation details/part of the inner working of the function.

apps/researcher/src/lib/user/actions.ts
```diff
@@ -0,0 +1,24 @@
...
+  const user = await clerkClient.users.getUser(userId);
+  if (!user) {
+    throw new Error(`User with ID ${userId} does not exist`);
+  }
+
+  let iri = user.unsafeMetadata?.iri as string | undefined;
+  if (!iri) {
+    iri = createPersistentIri();
+
+    await clerkClient.users.updateUserMetadata(userId, {
```
sdevalk, line 16: To be sure: do we need to handle an error from Clerk (e.g. could `updateUserMetadata()` return null or undefined if the update fails)? Or will Clerk simply throw an error if the update is unsuccessful?

sdevalk (approved):

PR #767, opened 2025-07-18
feat: inLanguage

sdevalk:
This PR filters the citations of research aids based on their language (e.g. Dutch or English). This is to prevent English citations appearing in a Dutch research aid, and vice versa.

As far as I can see we do not need to present the languages of citations in the frontend.

Source: https://github.com/colonial-heritage/sprints/issues/502


barbarah (approved):
Nice! I'm glad we have a solution for this.

PR #774, opened 2025-07-24
feat: order research aids

sdevalk:
This PR adds ordering to the 'member aids' (`hasParts`) of a research aid and to the 'related aids' (`seeAlso`) of a research aid.

It does so by introducing a new, optional property on a `ResearchGuide`: `position`. The frontend can use this property to sort the research aids. If the `position` of a research aid is `0`, then either it is the first aid in a list or the position was not set explicitly in the NIOD source data (in which case there can be more research aids with a position `0`).

**This PR, at this moment, only works with the data on the testing environment**, accessible via https://dev.app.colonialcollections.nl/nl/research-aids. Some changes still need to be made to the NIOD ETL before the data with the ordering is available on production.

The new data from NIOD has been put in a temporary 'acceptance' knowledge graph, so that we can test the changes to the data. This knowledge graph can be queried in the frontend by setting env var `SPARQL_ENDPOINT_URL` to `https://api.colonialcollections.nl/datasets/data-hub/knowledge-graph-acceptance/sparql`.

Source: https://github.com/colonial-heritage/sprints/issues/500

### DIFF
--- a/packages/api/src/research-guides/rdf-helpers.ts
+++ b/packages/api/src/research-guides/rdf-helpers.ts
@@ -82,11 +82,10 @@ function createResearchGuideFromListItem(
     getPropertyValues(listItemResource, 'ex:position')
   );
 
-  // If `position` is missing, use `0` instead
-  const position = rawPosition !== undefined ? parseInt(rawPosition) : 0;
-
-  // The position of the guide within the current list
-  researchGuide.position = position;
+  if (rawPosition !== undefined) {
+    // The position of the guide within the current list
+    researchGuide.position = parseInt(rawPosition);
+  }
 
   return researchGuide;
 }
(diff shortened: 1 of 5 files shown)


packages/api/src/research-guides/rdf-helpers.ts
```diff
@@ -78,11 +66,61 @@ export function createEvents(resource: Resource, propertyName: string) {
...
+  }
+
+  const researchGuide = createResearchGuide(researchGuideResource, stackSize);
+
+  const rawPosition = onlyOne(
+    getPropertyValues(listItemResource, 'ex:position')
+  );
+
+  // If `position` is missing, use `0` instead
+  const position = rawPosition !== undefined ? parseInt(rawPosition) : 0;
```
barbarah, line 86: I prefer the default to be undefined. 0 would place these items at the top of the list. And I can imagine that items without a position need to be at the back.

Or, if you prefer `0`, make `position` not optional in the TypeScript interface. So the frontend does not need to have logic for both zero and undefined.

sdevalk: That makes sense! I'll make the default 'undefined'.

barbarah (approved):

PR #778, opened 2025-08-01
Update messages.json

CamideK55:
small edit in frontend text


valevo (commented):
"Maker(s)" might also be more apt in English; there's no real consensus terminology about this but "maker" fits many contexts ("creator" gets into weird territory with conceptual creation)

valevo:
@CamideK55 -- shouldn't necessarily be mentioned here but because I just noticed it:
the change you made to makers, putting the plural into braces, could also apply to other fields such as "Techniques", "Materials", etc. definitely more important for makers but maybe for consistency. please feel free to make a decision

CamideK55:
@valevo , have a look, made an update. Only thing, linguistic wise, this is more difficult to do in Dutch - with the plurals

PR #781, opened 2025-08-01
Update messages.json

CamideK55:
plural edits in dutch


CamideK55:
@valevo , I know! haha, but I can't says 'material(en)' because its 'materiaal', so I would have to make it: 'materiaal/materialen', but Im not sure about that...

valevo:
I know, silly Dutch language, changing stems across inflections... (don't worry, other languages do this, too)
I think it's almost never just a single matieral, so the plural could make sense. but then please make it consistent with the Dutch version :)

CamideK55:
haha, mean : (. 
but making it consistent would mean using plural everywhere.
I would make it 'materiaal/materialen'

PR #785, opened 2025-08-18
Bump node to 22

jauco:
Vercel won't support Node 18 from sep 1 onwards. 

Tagging both Sjors and Barbara because I'm not sure who's on holiday

### DIFF
--- a/.github/workflows/database-migrate.yml
+++ b/.github/workflows/database-migrate.yml
@@ -2,8 +2,8 @@ name: Migrate database
 
 on:
   pull_request:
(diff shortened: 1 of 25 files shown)


barbarah (approved):
I have added two missing dependencies and fixed the error "TypeScript enum is not supported in strip-only mode".

https://github.com/microsoft/playwright/issues/34263

The end-to-end tests still fail due to an unrelated problem (see email sent to Gertjan and Sjors). I'm ok with merging this request.

PR #796, opened 2025-09-29
fix: make locale required in citations

sdevalk:
This PR fixes an issue in the display of the research aids: some sources of research aids only consist of URLs [1].

[1] Example: https://app.colonialcollections.nl/en/research-aids/https%3A%2F%2Fn2t%252Enet%2Fark%3A%2F27023%2F05a5c7a799263dcf493f0a7bdc2f525a

### DIFF
--- a/packages/api/src/research-guides/fetcher.integration.test.ts
+++ b/packages/api/src/research-guides/fetcher.integration.test.ts
@@ -22,11 +22,11 @@ describe('getTopLevels', () => {
           'Research aides for conducting provenance research into colonial collections',
         text: 'On this page you find various research aides that can assist...',
         encodingFormat: 'text/markdown',
-        hasParts: [
+        hasParts: expect.arrayContaining([
           {
             id: 'https://guides.example.org/subset1',
             name: 'Name',
-            hasParts: [
+            hasParts: expect.arrayContaining([
               {
                 id: 'https://guides.example.org/guide3',
                 name: 'Sources',
@@ -42,13 +42,13 @@ describe('getTopLevels', () => {
                 name: 'How can I use the data hub for my research?',
                 position: 5,
               },
-            ],
+            ]),
             position: 1,
           },
           {
             id: 'https://guides.example.org/subset3',
             name: 'Name',
-            hasParts: [
+            hasParts: expect.arrayContaining([
               {
                 id: 'https://guides.example.org/guide6',
                 name: 'Royal Cabinet of Curiosities',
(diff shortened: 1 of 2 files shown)


jauco:
Hmm, I'm not sure I get what you're saying.

In the page you link to I see 'https://rkd.nl/library/89731' among others without a description. Is that what you're referring to as the things that this fixes?

The code change (based on your description and what I gather from the code) would hide that entry?

Because if I look at the source for that url I see https://github.com/colonial-heritage/research-guides-dev/blob/bfc42cc0fe8b4b1db17b29b0af660fb0445d8b20/published/niveau3/English/KoninklijkeBazar_2040503.yml#L89 and it does seem to have a description in the source. So if we hide it then I don't think we fixed the issue from a user's perspective?

But I might just be missing what is going on here.

sdevalk:
@jauco Thanks!

Some detail pages of research aids show what look like 'orphaned' sources - sources consisting of just a URL. These sources shouldn't be here - they're not in the locale chosen by the user (English or Dutch). This is caused by the SPARQL query responsible for fetching the sources - the query contained an `OPTIONAL` statement, indicating that any source is valid, regardless of its locale, whereas only sources of the chosen locale are valid.

jauco:
Check. So what url will be filtered out by this change? And then the fact that https://rkd.nl/library/89731 does not have a description should be fixed by a different PR?

sdevalk:
In response:

> So what url will be filtered out by this change?

To be sure: it's not about URLs - it's about sources (primary or secondary). Sources (can) have URLs. This change aims to filter sources that are not available in the locale chosen by the user.

> And then the fact that https://rkd.nl/library/89731 does not have a description should be fixed by a different PR?

The current PR should fix this: the source with URL https://rkd.nl/library/89731 is not available in the chosen locale, but displayed nonetheless because of the erroneous SPARQL query. If it doesn't fix the issue, something else is going on.

jauco:
> The current PR should fix this: the source with URL https://rkd.nl/library/89731 is not available in the chosen locale

it is available in the locale, it's right there on the page you linked to

image 

and it's also on the page and line I linked to above.

It's a secondary source but its url is also listed under the primary sources somehow. So the locale does not seem to have anything to do with this?

sdevalk:
@jauco Thanks! In response:

> it is available in the locale, it's right there on the page you linked to

The [page of the research aid](https://app.colonialcollections.nl/en/research-aids/https%3A%2F%2Fn2t%252Enet%2Fark%3A%2F27023%2F05a5c7a799263dcf493f0a7bdc2f525a) shows two occurrences of the URL `https://rkd.nl/library/89731`. This URL is a part of two sources.

The first occurrence of the URL, listed in the section "Primary sources", belongs to [this source](https://n2t.net/ark:/27023/bd1491ceaf03af6c176913c98b516f82), "Artikel over de winkel van Dirk Boer uit het tijdschrift voor Japanse kunst.". This source is available in the Dutch locale (`inLanguage`: `nl`).

The second occurrence of the URL, listed in the section "Secondary sources", belongs to [this source](https://n2t.net/ark:/27023/1978465f110e3167f249215827bdd829), "Article about Dirk Boer's store from the Journal of Japanese Art.". This source is available in the English locale (`inLanguage`: `en`).

The source with the first occurrence of the URL shouldn't be displayed on the page: the page is in English whereas the source is in Dutch. The change to the SPARQL query in the PR aims to fix this: it did not properly check for the locale.

jauco:
ah, okay. Then what tripped me up was the fact that (a) you consider them different sources. I would have expected one source (one URI) with multiple triples for the description predicate on that source to define the various translations. It took me until this response to backtrack on that reasoning.

But it was reinforced by the fact that the 'source' that has the wrong locale is displayed under a different section. I'm assuming that is a fallback that erroneously kicks in?

Approving BTW. Though if you could add a small global description of how this works to the sparql query (or a pointer to a central document) that would improve the ability for a new dev to enter the codebase a lot!

jauco (approved):

PR #799, opened 2025-09-30
Clerk core update

barbarah:
# Migrate to Clerk Core 2 (v4→v6)

This PR completes the migration from Clerk v4 to v6, implementing Clerk Core 2 architecture. Core 1 / `@clerk/nextjs` v4 is no longer supported. See Clerk’s [Versioning & LTS](https://clerk.com/docs/upgrade-guides/long-term-support) page.

## Scope

This PR does not change product behavior.

Note on UI: the Community Profile modals are changed (see point 4 in Changes). This has a small styling impact. You can see these modals on a community page you are a member of: click “Edit members” and “Edit community” → “More settings”.

## Background problems Core 1

In **Core 1** we sometimes had an auth mismatch: the client showed a signed-in user, but the backend (`currentUser()` / `auth()`) returned `null`. This was hard to find and fix, so we moved more auth logic to the frontend to avoid the issue.

**Core 2** (Clerk v6) includes many fixes and improvements. This might solve the mismatch. My advice is to keep monitoring for any remaining issues before removing the auth health check.

### What I did not remove (but did update to core 2)

* **AuthHealthCheck**: this is our check for the issue above. It loads in the main layout and checks if auth is successful in both frontend and backend. It keeps logging when the frontend shows a user but the backend does not (or the other way around). The logs can be found in Vercel (retention: ~1 day).
* **Debug page**: `/debug`. This page helps when the mismatch happens. A user can open it and click the logging button. That sends a detailed trace (client state, server responses). It may be obsolete with Core 2, but I’m keeping it until we’re sure.
* **Frontend auth approach**: I did not move logic to the backend. If Core 2 is stable, we could make things less front-end heavy and rely a bit more on the backend.

## Changes

1. **Clerk Core 2 Migration (Automated + Manual Fixes)**

   * Updated `middleware.ts` to use `clerkMiddleware` instead of deprecated `authMiddleware`
   * Fixed server-side imports: `currentUser`, `auth`, and `clerkClient` from `@clerk/nextjs/server`
   * Updated API return types: `.data` property for `getOrganizationMembershipList` and `getOrganizationList`
   * Changed component props: `afterSignXUrl` to `fallbackRedirectUrl` in auth pages
   * Updated `clerkClient` to new async `await clerkClient()` pattern
   * Fixed test files to work with new Organization/OrganizationMembership types
   * Removed deprecated `session` property from auth destructuring

2. **Package Dependency Changes**

   * **Production Dependencies**:

     * Removed `@clerk/shared` package (was only used for error handling)
     * Updated error handling to use new Clerk Core 2 error patterns
   * **Development Dependencies**:

     * Added `@clerk/backend` 2.16.0 as dev dependency for latest API compatibility in e2e tests
     * Updated `@clerk/testing` from 1.1.5 to 1.13.0
     * Removed deprecated `@clerk/clerk-sdk-node` package

3. **Localization Updates**

   * **Step 1**: Replaced old custom translations with the [latest Clerk translations](https://github.com/clerk/javascript/blob/aaab21703a239d7bdd4866f9d56fd60cd337813c/packages/localizations/src/en-US.ts)

     * Updated `@clerk/localizations` from 1.26.13 to 3.25.3
     * Reset both English and Dutch files to the full official Clerk structure
   * **Step 2**: Applied organization → community wording:

     * English: “organization(s)” → “community/communities”
     * Dutch: “organisatie(s)” → “community/communities”
     * Fixed grammar: “Choose an community” → “Choose a community”
     * Kept case and plural forms

4. **Community Profile Modal Workaround**

   * **Background:** Clerk’s `OrganizationProfile` shows both general settings and member settings in one modal (when using the hook `openOrganizationProfile`) or component (when using `<OrganizationProfile />`).  We don’t want the full Clerk UI; we only wish to separate pages opened from our own buttons (“Edit members”, “Edit community” → “More settings”). So we had a workaround, using `openOrganizationProfile()` and hiding the part we didn't want to show.
   * **Problem:** In v6, `openOrganizationProfile()` doesn’t open a specific tab (members vs general).
   * **Old approach:** Call `openOrganizationProfile()` with `customPages` ordering and a hidden navbar.
   * **New approach:** Create our own modal, use place the `<OrganizationProfile />` component in the modal. The OrganizationProfile component accepts hash routing, allowing us to open the chosen tab with a hash.

     * Uses `window.location.hash` for tabs: `#/organization-members`, `#` (for general).
     * Forces `OrganizationProfile` to re-mount by changing a `key` when the hash updates
     * Hide the parts we don't want to see (navigation to other tabs)
   * **Files:**

     * `community-profile-modal.tsx` — modal with hash awareness and comments
     * `hooks.tsx` — `useCommunityProfile` updated to drive hash routing

5. **End-to-End Test Improvements**

   * **Infra fixes**:

     * Updated imports to `@clerk/backend` (removed `@clerk/clerk-sdk-node`)
     * Switched to `clerk.signIn()` helper
     * Better error logs + debug screenshots
   * **Reliability**:

     * Separate `deleteUser` / `deleteCommunity`
     * Better handling for community creation errors
     * Fixed assertion: “Create Community” → “Create community”
     * CI test script uses correct env handling


jauco:
Ik had afgelopen week een beetje een volle week, die nu ongeveer afgerond is. Ik zal deze vandaag reviewen

apps/researcher/e2e/lib/community.ts
```diff
@@ -16,15 +36,25 @@ export async function createUser({
...
+    return await clerkClient.users.createUser({
+      firstName,
+      lastName,
+      emailAddress: [emailAddress],
+      password,
+      unsafeMetadata: {
+        iri: `https://example.com/${testId}`,
+      },
+    });
+  } catch (error: any) {
```
jauco: Out of curiosity: Why the extra log wrapper?

apps/researcher/src/lib/community/actions.ts
```diff
@@ -93,27 +97,33 @@ export async function getMyCommunities({
...
   const memberships = userId
-    ? await clerkClient.users.getOrganizationMembershipList({
-        userId,
-        limit,
-        offset,
-      })
-    : [];
-
-  const organizations = memberships.map(membership => membership.organization);
+    ? await (async () => {
```
jauco: nit

```
 const memberships = userId
    ? await clerkClient().then(client => 
        client.users.getOrganizationMembershipList({
          userId,
          limit,
          offset,
        })
      )
     : {data: [], totalCount: 0};
```

apps/researcher/src/lib/community/actions.ts
```diff
@@ -93,27 +97,33 @@ export async function getMyCommunities({
...
+          const client = await clerkClient();
           const members =
-            await clerkClient.organizations.getOrganizationMembershipList({
+            await client.organizations.getOrganizationMembershipList({
               organizationId: organization.id,
             });
-          return organizationToCommunity({
-            ...organization,
-            membersCount: members.length,
+          const organizationWithCount = Object.assign({}, organization, {
```
jauco: OOC why no longer the spread?

apps/researcher/src/app/[locale]/communities/[slug]/community-profile-modal.tsx
```diff
@@ -0,0 +1,89 @@
...
+      window.removeEventListener('hashchange', handleHashChange);
+    };
+  }, []);
+
+  return (
+    <Modal id={modalId} variant="full">
+      <ModalHeader />
+      <div className="mt-4">
+        {/*
+          Key prop is CRITICAL: Forces OrganizationProfile to re-mount when hash changes.
```
Comment (jauco) [inline apps/researcher/src/app/[locale]/communities/[slug]/community-profile-modal.tsx:61]: 🙇

apps/researcher/src/lib/community/hooks.tsx
```diff
@@ -1,38 +1,30 @@
...
-          customPageA === firstPage ? -1 : customPageB === firstPage ? 1 : 0
-        )
-        .map(page => ({label: page})),
-      appearance: {
-        elements: {
-          navbar: 'hidden',
-        },
-      },
+
+    // CRITICAL: Set hash BEFORE opening modal to ensure correct initial tab
```
jauco:
```suggestion
    // CRITICAL: Set hash BEFORE opening modal so that you don't have a flash of the wrong tab
```

(I misunderstood this comment at first to mean that otherwise it would never end up on the right tab and was confused)

jauco (approved):
Some small questions/nits. LGTM!

Also, thanks for the elaborate description. Otherwise I would've had a lot more questions!

apps/researcher/e2e/lib/community.ts
barbarah: The try-catch wrapper was added during debugging and isn't actually needed. 

The `createUser` call will already throw an error if user creation fails, so wrapping it in try-catch just to log and re-throw the same error adds unnecessary complexity.

I've simplified it to call `createUser` directly.

apps/researcher/src/lib/community/actions.ts
barbarah: This was because of a typescript error, but there are better ways to solve this. I've switched back to the spread operator with a proper type cast:

```typescript
const organizationWithCount = {
  ...organization,
  membersCount: members.data.length,
} as typeof organization & {membersCount: number};

barbarah (commented):
I fixed the remarks. Camiel had one small UI remark. I will fix this on Monday and merge after the UI fix.

PR #800, opened 2025-10-02
feat(images): 30-day edge cache, Tailwind-aligned sizes, enable AVIF/WebP

barbarah:
**Problem**
Edge cache for optimized images is too short for our use. Thumbnails should keep working during brief origin issues.

**Solution**
Set a ~30 day cache floor on Vercel and limit generated widths to what we actually render. Enable modern formats to reduce bytes. Low risk change; could solve the issue. Let’s test if this is enough.

**What I changed**

* I set `minimumCacheTTL: 2_592_000` in `next.config.js` (about 30 days).
* I aligned `deviceSizes` with Tailwind 3.4.4: `[360, 640, 768, 1024, 1280, 1536]`.
* I added `imageSizes: [80, 90, 120, 160, 270, 360]` to match our UI.
* I enabled `formats: ['image/avif', 'image/webp']`.
* I updated `sizes` values:

  * HeritageObjectCard Large: `(min-width: 1024px) 360px, (min-width: 640px) 270px, 160px`; Small: `90px`.
  * HeritageObjectListItem: `80px` (matches `w-20`).
  * Gallery hero: (min-width: 768px) 40vw, 50vw; thumbs: (min-width: 768px) 10vw, 20vw.

**Costs**

* What we pay for on Vercel: image transformations (first time a size is generated), cache reads/writes, edge requests, data transfer.
* This PR changes cache duration only. TTL is now ~30 days (2,592,000 s). Sizes stay constrained but similar.

**Expectation**

* I expect no price increase. Longer TTL means fewer cache writes and fewer repeat transformations. Bandwidth per view stays similar or lower (AVIF/WebP).
* No plan upgrade required for this change.

### DIFF
--- a/apps/researcher/next.config.mjs
+++ b/apps/researcher/next.config.mjs
@@ -13,11 +13,11 @@ const nextConfig = {
     mdxRs: true,
   },
   images: {
-    remotePatterns: [
-      {
-        hostname: '*',
-      },
-    ],
+    remotePatterns: [{protocol: 'https', hostname: '**'}],
+    minimumCacheTTL: 2_592_000, // 30 days
+    deviceSizes: [360, 640, 768, 1024, 1280, 1536],
+    imageSizes: [80, 90, 120, 160, 270, 360],
+    formats: ['image/avif', 'image/webp'],
   },
   redirects: async () => [
     {
--- a/apps/researcher/next.config.mjs
+++ b/apps/researcher/next.config.mjs
@@ -13,7 +13,7 @@ const nextConfig = {
     mdxRs: true,
   },
   images: {
-    remotePatterns: [{protocol: 'https', hostname: '**'}],
+    remotePatterns: [{hostname: '**'}],
     minimumCacheTTL: 2_592_000, // 30 days
     deviceSizes: [360, 640, 768, 1024, 1280, 1536],
     imageSizes: [80, 90, 120, 160, 270, 360],
--- a/apps/researcher/next.config.mjs
+++ b/apps/researcher/next.config.mjs
@@ -12,9 +12,11 @@ const nextConfig = {
   experimental: {
     mdxRs: true,
   },
+  // https://vercel.com/docs/image-optimization
   images: {
     remotePatterns: [{hostname: '**'}],
-    minimumCacheTTL: 2_592_000, // 30 days
+    // https://nextjs.org/docs/app/api-reference/components/image#minimumcachettl
+    minimumCacheTTL: 31_536_000, // 1 year
     deviceSizes: [360, 640, 768, 1024, 1280, 1536],
     imageSizes: [80, 90, 120, 160, 270, 360],
     formats: ['image/avif', 'image/webp'],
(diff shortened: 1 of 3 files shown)


apps/researcher/next.config.mjs
```diff
@@ -13,11 +13,11 @@ const nextConfig = {
     mdxRs: true,
   },
   images: {
-    remotePatterns: [
-      {
-        hostname: '*',
-      },
-    ],
+    remotePatterns: [{protocol: 'https', hostname: '**'}],
```
barbarah: Note to self: I see I added `protocol: 'https'`, but I think there was a reason this was not added. Need to check this tomorrow.

apps/researcher/next.config.mjs
```diff
@@ -13,11 +13,11 @@ const nextConfig = {
     mdxRs: true,
   },
   images: {
-    remotePatterns: [
-      {
-        hostname: '*',
-      },
-    ],
+    remotePatterns: [{hostname: '**'}],
+    minimumCacheTTL: 2_592_000, // 30 days
```
jauco: Cool! 

 - can we up it even more? The cache header for the museum has a max-age of 31_536_000 (A year)
 - Is there a config option to just listen to their cache header? Or is this setting _our_ upper limit? In the latter case I would definitely make it a year at least.
 - Could you add a link to the relevant docs for vercel in a comment. It will help us a lot in 6 months when we need to revisit this because something changed somewhere.

jauco (approved):
Wouldn't have thought about the image-size changes. Thanks!

apps/researcher/next.config.mjs
barbarah: Good question about the museum’s Cache-Control. As far as I can tell, next/image doesn’t have a switch to "mirror the origin TTL." ~~The museum’s header applies only when Vercel fetches the source image, it doesn’t control how long our optimized variant (/_next/image?...) is cached at the edge.~~ I also felt that 30 days may be too short, so I’ll update minimumCacheTTL to 1 year to match the museum's TTL.

One note from related research for Gertjan’s questions: The edge cache is per PoP (region) and evictable, so a long TTL doesn’t guarantee it stays warm. A warm cache in NL won’t help a first request in Indonesia.

barbarah: Clarification from: https://nextjs.org/docs/app/api-reference/components/image#minimumcachettl

The expiration (or rather Max Age) of the optimized image is defined by either the minimumCacheTTL or the upstream image Cache-Control header, whichever is larger. 

I’ll still set minimumCacheTTL to 1 year to match.

barbarah: Question: Do all museum image endpoints (all sources) return a Cache-Control: max-age=… header?

barbarah:
The end-to-end tests are failing again on saving enrichments. I'm not sure why. I know Sjors looked at it (send an email?) and then it worked, but now enrichments aren't saving again.

This has nothing to do with this pull request.

PR #809, opened 2026-01-23
Fix too many related guides appearing

jauco:
Related guides appear twice, once direct using a seeAlso relation and once through an intermediary itemList relation so that they can be ordered.

However, a guide page can itself also be an itemList, so the direct seeAlso link might cause the ordered components of the page to also be included as if they are related guides


CamideK55 (approved):
Discussed wit @jauco on slack.

PR #811, opened 2026-04-16
feat: replace onboarding animation with walkthrough videos

barbarah:
The homepage "How does this platform work?" section shows a static GIF animation. The consortium has Vimeo walkthrough videos that explain the platform better than the animation does.

This pull request replaces the GIF with a tabbed video player that loads walkthrough content from YAML files per locale. Each video entry has an active flag, set to false to hide.

## What I changed

- I created a walkthrough/ component with a Zod schema that parses YAML, validates entries, and silently skips invalid or inactive ones (we don't want a broken homepage)
- I added walkthrough.yaml for both EN and NL with all 6 videos. Only the first 2 are active right now, the rest are set to active: false
- I swapped the GIF `<Image>` in page.tsx for the new `<Walkthrough>` component and kept the intro text
- I updated `howThisWorksText` in both locales to reference the videos instead of the old animation
- I added unit tests covering valid YAML, invalid entries, empty text, inactive entries, and edge cases
- bundle walkthrough YAML at build time. Source files aren't deployed to Vercel, so readFile fails at runtime. Import YAML as raw strings via webpack asset/source rule instead.

## Before merge

- Set the second video ("Searching objects and belongings") to `active: false` - I did not do this already, so the preview shows the design with more than one video.
- Review all text changes/additions.
- Camiel: Before activating any video, make sure it's set to public in Vimeo first. The second video currently doesn't load because it's still private

Preview: https://colonial-collections-researcher-git-videos-colonial-heritage.vercel.app/

### DIFF
--- a/apps/researcher/next.config.mjs
+++ b/apps/researcher/next.config.mjs
@@ -12,6 +12,13 @@ const nextConfig = {
   experimental: {
     mdxRs: true,
   },
+  webpack: config => {
+    config.module.rules.push({
+      test: /\.yaml$/,
+      type: 'asset/source',
+    });
+    return config;
+  },
   // https://vercel.com/docs/image-optimization
   images: {
     remotePatterns: [{hostname: '**'}],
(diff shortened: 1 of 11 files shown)


jauco:
Deferring to Camiel for final approval, but LGTM!

CamideK55 (commented):
Hi @barbarah / Barbara, veel dank!

- de tweede video heb ik naar active: false gezet
- de teksten heb ik hier en daar wat aangepast.

Wel zie ik dat sommige end-to-end test nu failen. Wil je hier nog even naar kijken?

Dank!

CamideK55 (approved):
Thanks, Barbara!
Ready.

